PR #13, opened 2018-11-11
Scripts for auditing Mozart Jobs

shitong01:
`clean_failed_s3_no_clobber_datasets_add_tag,py`
 - for cleaning failed no_clobber jobs in s3 bucket.
 - added job_type filter optionto specify job_type to clear
 - option to add tags to failed no_clobber jobs which has s3 cleared for requeue

`clean_and_audit_slings.py`
 - clears sling bottlenecks in "job-queue" in case of unsynchronised redis servers
 - checks all job-failed / job-queued slings, checks if dataset in GRQ
 - if dataset in GRQ/ already queued tag the job as deduped for purging later.
 - if dataset not in GRQ, tag job for deletion and clear in redis for qquery to deal with later

`clean_and_audit_redis_for_slings.py`
- Audits redis. Clear redis entries if incomng SLC.zip entry in redis is not found in GRQ.

`clean_and_audit_ifg.py`
 - does the same as `clean_failed_s3_no_clobber_datasets_add_tag.py` + checks if s3 no_clobber failure is because dataset is already in GRQ, add deduped tags for those which are already in GRQ

`submit_data_extract.py`
 - Audits data-extract user-rule for extracting SLCs. Scans all incoming SLC.zip and check if they have been extracted. If not extracted, submits extraction jobs.


pymonger:
@shitong01, these are great additions. Let's move these to a `scripts` directory in the `ariamh` repo.

PR #24, opened 2019-04-10
HC-28 Triage ID is now configureable

fgreg:
This PR allows the Triage ID to be configureable via a context parameter.

Note: This patch is backward-compatible with python2


fgreg:
I've held this in draft because I wanted to provided documentation along with the patch. But, documentation for this feature is going to rely on setting up ReadTheDocs. So, I'm going to hold off on writing the documentation until I can get that setup. I've opened https://hysds-core.atlassian.net/browse/HC-42 to track the status of the documentation.

PR #25, opened 2019-04-26
added pre-processing checksum functions in hysds/utils.py

DustinKLo:


hysds/utils.py
```diff
@@ -737,3 +737,143 @@ def mark_localized_datasets(job, ctx):
...
+
+        calculated_checksum = calculate_checksum_from_localized_file(file_path, algo)
+        pre_computed_checksum = read_checksum_file(file_path_checksum)
+
+        if calculated_checksum.lower() != pre_computed_checksum.lower():
+            mismatched_checksums.append(file_path)
+            exception_string += ('%s: calculated checksum: %s, pre-computed checksum: %s\n' % (file_path, calculated_checksum, pre_computed_checksum))
+
+    if len(mismatched_checksums) > 0:
+        raise Exception(exception_string)
```
pymonger: Return an explicit `True` if no mismatched checksums were found.

hysds/utils.py
```diff
@@ -737,3 +737,143 @@ def mark_localized_datasets(job, ctx):
...
+
+    mismatched_checksums = []
+    exception_string = 'Files with mismatched checksum:\n'
+
+    for file_info in files_to_validate:
+        algo = file_info['algo']
+        file_path_checksum = file_info['file_path']
+        # this has the hash extension to the file, we need to remove it
+        file_path = file_path_checksum.replace('.' + algo, '')
+
```
pymonger: You think we should check if the `file_path` exists here and skip if it doesn't?

hysds/utils.py
DustinKLo: Yes, i tested the functions individually in a local Docker container, also added unit test file `test/test_preprocessing_checksum.py`

hysds/utils.py
DustinKLo: noted

pymonger:
@DustinKLo, I've added the configuration needed to integrate circleci to run pull request building. Please update your clone of the `HC-34-checksum_validation` branch by running `git pull`. Next, if you click on the `Details` for the failed circleci test above, you'll see that the  test failed here:

```
==================================== ERRORS ====================================
_____________ ERROR collecting test/test_preprocessing_checksum.py _____________
ImportError while importing test module '/home/ops/project/test/test_preprocessing_checksum.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
test/test_preprocessing_checksum.py:10: in <module>
    from test import hashlib_mapper, calculate_checksum_from_localized_file, read_checksum_file, \
E   ImportError: cannot import name 'hashlib_mapper' from 'test' (/opt/conda/lib/python3.7/test/__init__.py)
```
Please fix your test so that the above check passes. All you need to do make sure it runs on your dev machine then push the commits up to the `HC-34-checksum_validation` branch. That should trigger circleci to rerun the build/test again.

DustinKLo:
@pymonger  noted, i didn't change the import statement and add `sys.path.insert(0, os.path.join(os.getcwd(), 'hysds'))` 
but the CircleCI tests pass now 👍

hookhua:
Looks good. key is to walk the localized work dir to find collocated "file.hash_ext" signatures.

To make it perform slightly faster, update def hashlib_mapper(algo) to do the cascade check with the most often used checksum first to make it short out faster. e.g. md5 first.

PR #35, opened 2019-10-26
hard coded _doc type in dataset_ingest.py

DustinKLo:
added more stuff to gitignore

### DIFF
--- a/hysds/dataset_ingest.py
+++ b/hysds/dataset_ingest.py
@@ -358,8 +358,8 @@ def ingest(objectid, dsets_file, grq_update_url, dataset_processed_queue,
     # get level
     level = r.getLevel()
 
-    # get type
-    dtype = r.getType()
+    # get type (not needed since ES7+ uses only _doc type)
+    # dtype = r.getType()
 
     # set product metrics
     prod_metrics = {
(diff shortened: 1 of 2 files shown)


hysds/recognize.py
```diff
@@ -136,14 +136,6 @@ def getLevel(self):
         else:
             return self.recognized['level'].format(**self.group_dict)
 
-    def getType(self):
```
pymonger: @DustinKLo: there may be some confusion here. This function returns the dataset type as defined in the `datasets.json`, e.g. https://github.com/hysds/hysds/blob/develop/configs/datasets/datasets.json#L7. So this should remain there.

hysds/dataset_ingest.py
```diff
@@ -604,7 +603,7 @@ def ingest(objectid, dsets_file, grq_update_url, dataset_processed_queue,
         'ipath': ipath,
         'system_version': version,
         'dataset_level': level,
-        'dataset_type': dtype,
```
pymonger: @DustinKLo: let's preserve `dataset_type` to be what is specified in the `datasets.json` for the respective recognized dataset e.g. https://github.com/hysds/hysds/blob/develop/configs/datasets/datasets.json#L7.

DustinKLo:
i dont think i even need to make changes to this part because the actual indexing of the dataset is handled in the grq2 rest API

PR #46, opened 2020-01-17
Hc 153 logstash

DustinKLo:


DustinKLo:
`configs/logstash/indexer.conf.mozart`
to get float timestamps through logstash, i had to:
1. convert the value to string
2. create new field that splits the string by `.`
3. delete the old fields `timestamp` and `local_received`
4. rename the new fields to the old naming convention

DustinKLo:
I'm not sure how to fix the circleci test fail

pymonger:
@DustinKLo: we need to ensure that the right version of the dependencies are installed. Try pushing this up to your branch:
```
(verdi) ops@0c04296296a2:~/verdi/ops/hysds$ git diff
diff --git a/.circleci/config.yml b/.circleci/config.yml
index d171fa3..8e97adb 100644
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,6 +9,7 @@ jobs:
           name: Test
           command: |
             source $HOME/.bash_profile
+            pip install -e .
             cp configs/celery/celeryconfig.py.tmpl celeryconfig.py
             python setup.py test
 ```

**Comment by DustinKLo** [comment]
branch not ready to merge, getting a `400` ``block in setup_after_successful_connection'` error when logstash attempts to install the templates to elasticsearch

**Comment by DustinKLo** [comment]
@pymonger  fixed the issue, the `template` field got renamed to `index_patterns` in ES 7
so the 400 error was accurate

File: hysds/__init__.py [hunk 1]
```diff
@@ -2,6 +2,6 @@
 from __future__ import print_function
 from __future__ import division
 from __future__ import absolute_import
-__version__ = "0.3.5"
+__version__ = "0.3.6"
```
pymonger: @DustinKLo: let's shoot for version `0.4.0` instead and leave the `0.3.*` for the ES 1.7 branches.

PR #49, opened 2020-03-06
ping msgpack; version 1.0.0 introduces breaking changes

pymonger:
@DustinKLo: Please review. In order for development of HySDS core v3 to continue concurrent to HySDS core v4 development, we need to pin msgpack in `develop` since `develop` is technically still HySDS core v3 (logstash 1.5.5). This fix isn't needed in `develop-es7` branch but should be merged in from `develop` once this PR is merged so as to maintain the branch history (i.e. when we finally merge `develop-es7` into `develop` we won't have to worry about resolving this conflict).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ from setuptools import setup, find_packages
 import hysds
 
 setup(
-    name='hysds',
+    name="hysds",
     version=hysds.__version__,
     long_description=hysds.__description__,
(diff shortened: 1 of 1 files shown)


DustinKLo:
LGTM 👍 
i'll merge it into `develop-es7` and remove the pin once we iron out `v4.0.0`

PR #52, opened 2020-03-30
Develop es7

DustinKLo:


.circleci/config.yml
```diff
@@ -9,6 +9,22 @@ jobs:
           name: Test
           command: |
             source $HOME/.bash_profile
+            pip install -e .
```
pymonger: @DustinKLo: once this is in the `develop` branch and hysds_commons has merged `develop-es7` into `develop`, can we remove this block entirely?

DustinKLo: sounds good

DustinKLo: i'll merge the PR, then create another PR with the new `.circleci` file

PR #58, opened 2020-04-12
Hc 169

DustinKLo:
@mcayanan  this should fix the error, how do you test this?

### DIFF
--- a/hysds/job_worker.py
+++ b/hysds/job_worker.py
@@ -24,6 +24,7 @@ import re
 import shlex
 import signal
 from datetime import datetime
+from itertools import compress
 from subprocess import check_output, CalledProcessError
 from celery.exceptions import SoftTimeLimitExceeded
 from celery.signals import task_revoked


mcayanan:
@DustinKLo Good question. The way I ran into this error initially was that an exception was thrown during a custom post process we have on swot. Default post processes of HySDS are to publish a dataset to ES or do triage. So basically you would need to force either of those to fail somehow. Also, we need to verify that those post processes return `False` upon a failure.

DustinKLo:
> @DustinKLo Good question. The way I ran into this error initially was that an exception was thrown during a custom post process we have on swot. Default post processes of HySDS are to publish a dataset to ES or do triage. So basically you would need to force either of those to fail somehow. Also, we need to verify that those post processes return `False` upon a failure.

@mcayanan since i don't have the custom post processors in my cluster can you pull my branch and `sds update` your SWOT cluster?
this change should work i just wanted to make sure

mcayanan:
@DustinKLo I can do that, but I don't think I'll be able to get to it until later this week.

pymonger:
@mcayanan: we're going to merge this PR so when SWOT runs the test please let us know how it works for you.

mcayanan:
@pymonger @DustinKLo Sounds good. We're in the middle of subsystem testing and do not anticipate switching over to v4 until middle/late summer if I had to guess. In the meantime, I've set up a parallel develop-es7 to keep up. If I'm able to get some free time, I'll definitely test this out and let you know. Thanks!

PR #64, opened 2020-05-02
Hc 208

DustinKLo:
https://hysds-core.atlassian.net/browse/HC-208


DustinKLo:
had to fix the jinja templating in case `~/.sds/config` doesnt have a `GRQ_ES_PORT`, `GRQ_AWS_ES` or `GRQ_ES_PROTOCOL` value

PR #65, opened 2020-05-06
Aria 17: watchdog for mozart job timedout

mkarim2017:
https://jira.jpl.nasa.gov/browse/ARIA-17
Using time limit from job spec for alerts


scripts/check_job_periodicity.py
```diff
@@ -186,7 +190,8 @@ def do_job_query(url, job_type, job_status):
         "sort": [{"job.job_info.time_end": {"order": "desc"}}],
         "_source": ["job_id", "payload_id", "payload_hash", "uuid",
                     "job.job_info.time_queued", "job.job_info.time_start",
-                    "job.job_info.time_end", "error", "traceback"],
+                    "job.job_info.time_end", "job.job_info.time_limit",
+                    "error", "traceback"],
         "size": 1
     }
     logging.info("query: %s" % json.dumps(query, indent=2, sort_keys=True))
```
pymonger, line 197: @mkarim2017: while you're updating these scripts, could you also update the way we query ES in them by utilizing the common ES library. Essentially we're getting rid of the use of `requests` to query ES as we encounter them throughout HySDS core repos. For example, https://github.com/hysds/hysds/blob/30611dced6f7788a6915b16344276a238c575f86/hysds/event_processors.py#L84.

scripts/check_queue_periodicity.py
```diff
@@ -193,7 +196,7 @@ def do_queue_query(url, queue_name):
         },
         "sort": [{"job.job_info.time_start": {"order": "desc"}}],
         "_source": ["job_id", "status", "job_queue", "payload_id", "payload_hash", "uuid",
-                    "job.job_info.time_queued", "job.job_info.time_start",
+                    "job.job_info.time_queued", "job.job_info.time_start", "job.job_info.time_limit",
                     "job.job_info.time_end", "error", "traceback"],
         "size": 1
     }
```
pymonger, line 202: @mkarim2017: same thing here. Let's replace `requests` ES querying.

scripts/check_job_periodicity.py
mkarim2017: Updated for ES query. Kept for the slack put.

scripts/check_queue_periodicity.py
```diff
@@ -193,21 +198,15 @@ def do_queue_query(url, queue_name):
...
-    url_tmpl = "{}/job_status-current/_search"
-    r = requests.post(url_tmpl.format(url), data=json.dumps(query))
-    if r.status_code != 200:
-        logging.error("Failed to query ES. Got status code %d:\n%s" %
-                      (r.status_code, json.dumps(query, indent=2)))
-    r.raise_for_status()
-    result = r.json()
-
+    ES = elasticsearch.Elasticsearch(url)
+    result = ES.search(index="job_status-current", body=json.dumps(query))
```
DustinKLo, line 209: @mkarim2017 you should use this connection instead: https://github.com/hysds/hysds/blob/develop/hysds/es_util.py

so that in case we move mozart to AWS Elasticsearch, we dont have to refactor this code

example: https://github.com/hysds/hysds/blob/develop/hysds/user_rules_job.py#L16-L28

scripts/check_queue_periodicity.py
mkarim2017: updated

scripts/check_queue_periodicity.py
mkarim2017: I had to put it in. a  try except clause as not everything that is required may not be there.
we can get error: 
ModuleNotFoundError: No module named 'aws_requests_auth'ModuleNotFoundError: No module named 'aws_requests_auth'

PR #73, opened 2020-05-27
ARIA-17: watchdog for mozart job timedout

mkarim2017:
git push --set-upstream origin ARIA-17-es7


DustinKLo:
a lot of formatting issues and unused variables (etc. `url`), but `develop-es1` is deprecated for the most part so it LGTM 👍

PR #85, opened 2020-09-03
HC-242: Resolved file name too long issue

mcayanan:
This PR fixes an issue where essentially if you keep re-running a job on a triage dataset, you'll eventually get a `[Errno 36] File name too long` as the triage dataset name will keep getting longer for each run. This PR addresses the issue by adding a `task` identifier to the default triage dataset name: 

`triaged_job-{job_id}_task-{job[task_id]}`

Then, in the triage function, logic was added to check for a triage dataset name using the following regex:

`triaged_job-(?P<job_id>.+)_task-(?P<task_id>[-\\w])`

If we get a match, then the `job_id` is used as the job identifier instead of using the default `job["job_info"]["id"]`

The following screenshot shows the result of running a job on a triaged dataset. This demonstrates that all we're doing is updating the task id in the triaged dataset name:

triaged-datasets-created

A unit test was also added to test this feature.

### DIFF
--- a/hysds/utils.py
+++ b/hysds/utils.py
@@ -969,12 +969,3 @@ def validate_checksum_files(job, cxt):
     else:
         logger.info("checksum preprocessing completed successfully")
     return True
-
-if __name__ == '__main__':
-    with open("_job.json", "r") as f:
-        job = json.load(f)
-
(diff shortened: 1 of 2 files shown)


pymonger (approved):
LGTM

PR #86, opened 2020-09-11
HC-269: integration SDSWatch into cluster deployment

pymonger:
This PR includes updates to integrate SDSWatch into HySDS core and cluster provisioning. The assumption is that we re-using the ELK stack installed on the metrics component/instance.

These updates were tested on a NISAR PCM dev cluster and end-to-end SDSWatch functionality verified according to https://hysds-core.atlassian.net/wiki/spaces/HYS/pages/548634892/SDSWatch#In-case-all-the-files-are-lost%2C-here-is-how-to-set-it-up.


anaerobia (approved):
Looks correct

PR #90, opened 2020-10-01
HC-280: Update RFC_6598_RE to support a bigger range

mcayanan:
Update the RFC_6598_RE regular expression to support a bigger IP range, specifically 100.64.x.x - 100.127.x.x

Tested it out on my local cluster on swot and it checked out fine.


DustinKLo:
LGTM, but im curious if we'll ever have ips with 3 digits in the IP
ie. 100.64.xxx.xxx vs 100.644.xxx.xxx

pymonger (approved):
That's one sweet regex!

mcayanan:
@DustinKLo I don't think the regex allows 100.644.x.x Is that what your concern was?

Screen Shot 2020-10-01 at 3 26 27 PM

DustinKLo:
yeah that was my concern, but besides that the changes LGTM 👍

PR #91, opened 2020-10-05
HC-281: authenticate docker hub pulls

pymonger:
Created contexts `docker-hub-creds` and `git-oauth-token` and updated CircleCI config for authenticated docker pulls.


mcayanan (approved):
LGTM. Thanks for the quick update!

PR #95, opened 2021-01-05
HC-310: Update call to psutil's disk_partitions function

mcayanan:
This PR updates the call to psutil's disk_partitions function due to API changes recently made. See the comment on the associated Jira ticket for evidence that this change resolves the issue.

### DIFF
--- a/scripts/log_instance_stats.py
+++ b/scripts/log_instance_stats.py
@@ -76,12 +76,14 @@ def daemon(redis_url, redis_key, interval):
             "disk_io": psutil.disk_io_counters()._asdict(),
             "net_io": psutil.net_io_counters()._asdict(),
         }
-        for device, mnt_point, fs_type, fs_opts in psutil.disk_partitions():
+        for device, mnt_point, fs_type, fs_opts, max_file, max_path in psutil.disk_partitions():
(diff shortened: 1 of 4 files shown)


scripts/log_instance_stats.py
```diff
@@ -76,12 +76,14 @@ def daemon(redis_url, redis_key, interval):
             "disk_io": psutil.disk_io_counters()._asdict(),
             "net_io": psutil.net_io_counters()._asdict(),
         }
-        for device, mnt_point, fs_type, fs_opts in psutil.disk_partitions():
+        for device, mnt_point, fs_type, fs_opts, max_file, max_path in psutil.disk_partitions():
```
DustinKLo, line 79: should we add `*other` in case they add more return values in `psutil.disk_partitions()`
`*other` can be used as a fail-safe and any new values will be set in a list

mcayanan: I guess it shouldn't hurt to add that in there. @pymonger any thoughts on this?

pymonger: Sounds good. It shouldn't hurt.

pymonger (approved):

PR #97, opened 2021-02-01
HC-300: improvements to core systemd services: harikiri and spot_termination_detector

mcayanan:
Updated the existing harikiri and spot_termination_detector scripts to support the passing in of a configuration file to read tool arguments from. The user will still have the ability to pass in arguments through the command line. Anything specified on the command line will take precedence over whatever gets specified in the configuration file. See the comments in the associated Jira ticket of how it was tested.

### DIFF
--- a/scripts/harikiri.py
+++ b/scripts/harikiri.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 """
 HySDS inactivity daemon to perform scale down of auto scaling group/spot fleet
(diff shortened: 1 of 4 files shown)


DustinKLo:
LGTM for the most part 🤷 i don't know too much about `harikiri` so i'll defer this to @pymonger 

I see a lot of empty `Exception` handling, should we log those errors?
```
try:
   ...
except Exception as e:
   logger.error(e)
```

pymonger (approved):
LGTM

PR #101, opened 2021-02-24
HC-334: Improve dedup detection

mcayanan:
This PR provides better support for detecting dedup jobs in the system. When the `query_dedup_job` function is called, it will first try to store the payload hash in REDIS. If it detects that it already exists, we will perform exponential backoff when trying to find the dedup job. This allows us to try and catch jobs with the same payload that may have been submitted at the same time. On the caller side, specifically at the orchestrator and job_worker levels, the calling of the `query_dedup_job` function is now surrounded by a try/except block to catch `NoDedupJobFoundException` exceptions so we can set the result to None if the hash key exists in REDIS and we cannot find the dedup job at the end of the exponential backoff.

This feature was tested on a local cluster, where this change was checked out onto Factotum and jobs were submitted to there for testing purposes. This screenshot shows that we are now able to properly catch the dedup even though the SCIFLO jobs were submitted at roughly the same time:

image

As an additional sanity check, this feature is being run through an end-to-end test on the SWOT side, where this issue initially appeared: https://swot-pcm-ci.jpl.nasa.gov/job/force-branches/job/ci-E2E_NIGHTLY-swot-pcm_develop-force-branch/308/

### DIFF
--- a/hysds/utils.py
+++ b/hysds/utils.py
@@ -343,7 +343,7 @@ def get_payload_hash(payload):
     ).hexdigest()
 
 
-def no_dedup_job():
+def no_dedup_job(e):
     return None
 
 
--- a/hysds/utils.py
+++ b/hysds/utils.py
@@ -47,6 +47,12 @@ grq_es = get_grq_es()
 DU_CALC = {"GB": 1024 ** 3, "MB": 1024 ** 2, "KB": 1024}
 
 
+class NoDedupJobFoundException(Exception):
+    def __init__(self, message):
+        self.message = message
+        super(NoDedupJobFoundException, self).__init__(message)
+
+
 def get_module(m):
     """Import module and return."""
 
@@ -352,7 +358,7 @@ def no_dedup_job(details):
     backoff.expo, requests.exceptions.RequestException, max_tries=8, max_value=32
 )
 @backoff.on_exception(
-    backoff.expo, ValueError, max_tries=8, max_value=32, on_giveup=no_dedup_job
+    backoff.expo, NoDedupJobFoundException, max_tries=8, max_value=32, on_giveup=no_dedup_job
 )
 def query_dedup_job(dedup_key, filter_id=None, states=None):
     """
@@ -411,7 +417,7 @@ def query_dedup_job(dedup_key, filter_id=None, states=None):
     logger.info("result: %s" % r.text)
     if j["hits"]["total"]["value"] == 0:
         if hash_exists_in_redis is True:
-            raise ValueError("Could not find any dedup jobs with the following query: {}".format(
+            raise NoDedupJobFoundException("Could not find any dedup jobs with the following query: {}".format(
                 json.dumps(query, indent=2)))
         elif hash_exists_in_redis is False:
             return None
--- a/hysds/utils.py
+++ b/hysds/utils.py
@@ -35,7 +35,7 @@ from atomicwrites import atomic_write
 from bisect import insort
 
 import hysds
-from hysds.log_utils import logger, log_prov_es
+from hysds.log_utils import logger, log_prov_es, payload_hash_exists
 from hysds.celery import app
 from hysds.es_util import get_grq_es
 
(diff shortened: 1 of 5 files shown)


hysds/utils.py
```diff
@@ -343,15 +349,29 @@ def get_payload_hash(payload):
...
+def no_dedup_job(details):
+    logger.info("Giving up querying for dedup jobs with args {args} and kwargs {kwargs}".format(**details))
+    return None
+
+
 @backoff.on_exception(
     backoff.expo, requests.exceptions.RequestException, max_tries=8, max_value=32
 )
+@backoff.on_exception(
+    backoff.expo, NoDedupJobFoundException, max_tries=8, max_value=32, on_giveup=no_dedup_job
```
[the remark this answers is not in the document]
mcayanan: My understanding of the exponential backoff is that multiple decorators are supported (https://pypi.org/project/backoff/):

image

Given this, it seems like the decorators here will handle both exceptions? Let me know if you think otherwise. 
With the `on_giveup` function, during testing I discovered that it will still re-raise the exception back to the caller even though the `no_dedup_job` returns None. That's the reason why the orchestrator and the job_worker needed to be updated as well to surround the `query_dedup_job` function with a try/except block to catch the NoDedupJobFoundException to essentially return None. If we decide to move forward with what you've suggested and combine the 2 exceptions into 1 decorator, I think the `on_giveup` function can be left as-is. Thoughts?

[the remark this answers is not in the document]
pymonger: @mcayanan: I like that approach better then. So the original decorator for handling `requests.exceptions.RequestException` should be restored and a second decorator handling the `NoDedupJobFoundException` added as you have here. Did I get that right? If so, let's go for it.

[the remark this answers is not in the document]
mcayanan: Cool. I think that's how it was implemented initially? Here's the change:

https://github.com/hysds/hysds/blob/ff62cd5e497b3c3b5117f423e3ae31b34184a269/hysds/utils.py#L357-L363

hysds/utils.py
```diff
@@ -343,15 +349,29 @@ def get_payload_hash(payload):
...
+)
 def query_dedup_job(dedup_key, filter_id=None, states=None):
     """
     Return job IDs with matching dedup key defined in states
     'job-queued', 'job-started', 'job-completed', by default.
     """
 
+    hash_exists_in_redis = payload_hash_exists(dedup_key)
```
DustinKLo, line 369: @mcayanan should we move this to line 418 inside this `if` statement?
```python
if j["hits"]["total"]["value"] == 0:
```
it can save a little bit of time of having to check redis `payload_hash` for every job

DustinKLo:
during the large update from ES1.7 -> 7.1 i forgot to convert the ES query in `query_dedup_job()` to use `elasticsearch-py`
https://github.com/hysds/hysds/blob/develop/hysds/utils.py#L386-L396
```python
    headers = {"Content-Type": "application/json"}
    r = requests.post(es_url, data=json.dumps(query), headers=headers)
    if r.status_code != 200:
        if r.status_code == 404:
            logger.info(
                "status_code 404, job_status-current index probably does not exist, returning None"
            )
            return None
        else:
            r.raise_for_status()
    j = r.json()
```
should we tackle this in this PR, or we can defer it to some other time

hysds/utils.py
pymonger: @DustinKLo: We need that atomic redis check up front to resolve the race condition between two duplicate jobs starting around the same time. Without that, they both will query ES and get the same result: that there are no other jobs in the system with that same payload hash.

pymonger:
> during the large update from ES1.7 -> 7.1 i forgot to convert the ES query in `query_dedup_job()` to use `elasticsearch-py`
> https://github.com/hysds/hysds/blob/develop/hysds/utils.py#L386-L396
> 
> ```python
>     headers = {"Content-Type": "application/json"}
>     r = requests.post(es_url, data=json.dumps(query), headers=headers)
>     if r.status_code != 200:
>         if r.status_code == 404:
>             logger.info(
>                 "status_code 404, job_status-current index probably does not exist, returning None"
>             )
>             return None
>         else:
>             r.raise_for_status()
>     j = r.json()
> ```
> 
> should we tackle this in this PR, or we can defer it to some other time

@DustinKLo: good catch. Let's defer that fix. Could you create a quick jira ticket for that?

pymonger (approved):

hysds/utils.py
DustinKLo: but it looks like we're not doing anything with the `hash_exists_in_redis` value until after we query elsaticsearch and nothing shows up?

mcayanan: When `hash_exists_in_redis` is True, that means we have identified that there is a possible duplicate job in the system. That's why in line 419, we do a check if this flag is True or not. If it is True, then we need to sort of "wait" and query ES a couple of times to give it time for the duplicate job to make it into the system.

DustinKLo (approved):
LGTM 👍

PR #102, opened 2021-02-24
HC-331: detect context JSON parameter to override the default no-clobber behavior of dataset publishing

pymonger:
The related Jira ticket is at https://hysds-core.atlassian.net/browse/HC-331.

This PR adds the capability for HySDS jobs to modify the `_context.json` within its job directory to add a `"_force_ingest": true` which is then propagated to the `publish_datasets` post processor so that the `hysds.dataset_ingest.ingest()` function is called with the `force=True` flag. With this parameter set, ingest of products that were already ingested will be overwritten and the job will not fail with a `no-clobber` exception. NOTE that this feature is contrary to the "data immutability" tenet of HySDS however it is understandable that this can help during development and testing of HySDS adaptation. 

2 new unit tests were also added to test this feature.

Documentation of this feature added here: https://hysds-core.atlassian.net/wiki/spaces/HYS/pages/199885846/Preprocessor+and+Postprocessor+Functions#Can-we-disable-this-feature-so-that-products-can-clobber-previously-ingested-products%3F.


mcayanan (approved):
LGTM. It looks like at the SWOT and NISAR level, we need to add a mechanism in there that will allow the operator to set the `_force_ingest` flag to True.

pymonger:
> LGTM. It looks like at the SWOT and NISAR level, we need to add a mechanism in there that will allow the operator to set the `_force_ingest` flag to True.

That's correct, @mcayanan. What Lela wants for I&T is a setting in the nisar-pcm `settings.yaml` that specifies whether or not specific PGEs in nisar-pcm should force ingest. If she sets, for example, INGEST_STAGED, to force ingest, then she'd have to run `sds ship` to reship out the updated `settings.yaml` and there would be a new chimera post-condition (or pre-condition, not sure which one) that will essentially modify the `_context.json` in the work dir and set this flag to true. Then verdi will force the ingest of the product to be ingested. That will come in the PR for nisar-pcm related to this ticket (NSDS-1199).

PR #116, opened 2021-05-25
HC-363: catch errors copying additional paths for triage

pymonger:
This PR adds more robust exception handling in the `triage` post-processor function and prevents jobs from failing when there are issues accessing additional file/directory globs. A new unit test was also added to test this feature.

### DIFF
--- a/hysds/__init__.py
+++ b/hysds/__init__.py
@@ -3,6 +3,6 @@ from __future__ import print_function
 from __future__ import division
 from __future__ import absolute_import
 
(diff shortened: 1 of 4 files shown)


mcayanan (approved):
LGTM. Thanks for the quick fix.

pymonger:
@mcayanan: sorry pushed one more update for HC-358. Can you re-review?

mcayanan (approved):
Ahh. Good catch. Ya that was a change we had to make as well.

PR #118, opened 2021-06-03
HC-374: failed job not detected as done causing harikiri to leave worker instances up

pymonger:
This PR fixes the issue where the `.running` signal file is not transitioned to `.done` because the remote services running on remote servers (e.g. ES, redis, or rabbitmq-server running on mozart) is not reachable. By transitioning the signal file upfront before these remote calls, the harikiri process on each verdi worker will be able to detect all jobs as being done and terminate the instance accordingly.

### DIFF
--- a/hysds/job_worker.py
+++ b/hysds/job_worker.py
@@ -1457,16 +1457,16 @@ def run_job(job, queue_when_finished=True):
 
     # close up job execution
     try:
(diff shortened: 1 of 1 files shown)


mcayanan (approved):
LGTM

DustinKLo (approved):

PR #122, opened 2021-07-01
HC-302: Add capability to propagate info messages up to the UI for successful jobs

mcayanan:
This PR adds the capability such that informational messages can get propagated up to the UI. This is useful for successful jobs that do not produce any output. 2 new fields are added to the job_status json:

- msg: Intended for short messages much like the error field. Can be a string or an array of strings.
- msg_details: Contains a detailed informational message.

If the worker finds an `_alt_msg.txt` file, it will add it to the `msg` field in the job_status. Additionally, if the `_alt_msg_details.txt` file is found, then it will add the string value to the `msg_details` in the job_status.

### DIFF
--- a/hysds/job_worker.py
+++ b/hysds/job_worker.py
@@ -1451,6 +1451,7 @@ def run_job(job, queue_when_finished=True):
         job_status_json["msg"] = msg
 
     # overwrite msg_details if _alt_msg_details.txt was dumped
+    msg_details = None
     alt_msg_details_file = os.path.join(job_dir, "_alt_msg_details.txt")
     if os.path.exists(alt_msg_details_file):
         with open(alt_msg_details_file) as f:
--- a/hysds/job_worker.py
+++ b/hysds/job_worker.py
@@ -1438,6 +1438,22 @@ def run_job(job, queue_when_finished=True):
             "celery_hostname": run_job.request.hostname,
         }
 
+    # overwrite message if _alt_msg.txt was dumped
+    alt_msg_file = os.path.join(job_dir, "_alt_msg.txt")
+    if os.path.exists(alt_msg_file):
+        with open(alt_msg_file) as f:
+            msg = f.read()
+            logger.info("Got alternate info message: %s" % msg)
+        job_status_json["msg"] = msg
+
+    # overwrite msg_details if _alt_msg_details.txt was dumped
+    alt_msg_details_file = os.path.join(job_dir, "_alt_msg_details.txt")
+    if os.path.exists(alt_msg_details_file):
+        with open(alt_msg_details_file) as f:
+            msg_details = f.read()
+            logger.info("Got alternate message details: %s" % msg_details)
+        job_status_json["msg_details"] = msg_details
+
     # store usage stats
     for usage_stats_file in find_usage_stats(job_dir):
         usage_stats = {}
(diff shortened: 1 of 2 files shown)


hysds/job_worker.py
```diff
@@ -1438,6 +1438,26 @@ def handler(signum, frame):
...
         }
 
+    # overwrite message if _alt_msg.txt was dumped
+    msg = list()
+    alt_msg_file = os.path.join(job_dir, "_alt_msg.txt")
+    if os.path.exists(alt_msg_file):
+        with open(alt_msg_file) as f:
+            msgs = f.readlines()
+            logger.info("Got alternate info message: %s" % msgs)
+        for m in msgs:
```
DustinKLo, line 1448: should the `for` loop be in the `with` block?

mcayanan: cool. thanks. done.

hysds/job_worker.py
```diff
@@ -1438,6 +1438,26 @@ def handler(signum, frame):
             "celery_hostname": run_job.request.hostname,
         }
 
+    # overwrite message if _alt_msg.txt was dumped
+    msg = list()
+    alt_msg_file = os.path.join(job_dir, "_alt_msg.txt")
+    if os.path.exists(alt_msg_file):
+        with open(alt_msg_file) as f:
+            msgs = f.readlines()
```
DustinKLo, line 1446: in the future should we expect users to write a ton of lines to `_alt_msg.txt`? (like thousands of lines per job)
should we change this to an iterator instead?

mcayanan: Its basically implementation specific. For `msgs`, it normally should only be 1 line, similar to how the `_alt_error.txt` is created. As we've seen in the Track Frame Evaluator, there could be more than 1 informational short message, but in that case it could go maybe as high as 3 lines.

hysds/job_worker.py
```diff
@@ -1438,6 +1438,26 @@ def handler(signum, frame):
...
+            logger.info("Got alternate info message: %s" % msgs)
+            for m in msgs:
+                msg.append(get_short_error(m))
+
+        job_status_json["msg"] = msg
+
+    # overwrite msg_details if _alt_msg_details.txt was dumped
+    alt_msg_details_file = os.path.join(job_dir, "_alt_msg_details.txt")
```
pymonger, line 1454:
```suggestion
    msg_details = None
    alt_msg_details_file = os.path.join(job_dir, "_alt_msg_details.txt")
```

pymonger: We should probably initialize `msg_details` so that line 1512 doesn't croak with an undefined variable error?

hysds/job_worker.py
pymonger: Good question. It is implementation specific but the best practice would be to use `_alt_msg.txt` for high-level, generic messages (probably no more than 5) and to use `_alt_msg_details.txt` for details on those generic messages.

hysds/job_worker.py
mcayanan: Thanks for catching that @pymonger. I've updated the code per your suggestion.

DustinKLo (approved):
LGTM 👍 but ill wait for @pymonger 's review

pymonger (approved):
LGTM too

PR #160, opened 2023-04-18
HOTFIX: forgot to raise error when async_publish_files fails

DustinKLo:
b/c we're not properly raising the error, it only logs it and technically the job will complete

### DIFF
--- a/hysds/dataset_ingest_bulk.py
+++ b/hysds/dataset_ingest_bulk.py
@@ -875,11 +875,12 @@ def async_publish_files(job, ctx, prod_dir, event=None):
         }
 
         return prod_json, product_staged_metadata, metrics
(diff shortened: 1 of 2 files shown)


pymonger (approved):

hysds/dataset_ingest_bulk.py
```diff
@@ -875,11 +875,12 @@ def async_publish_files(job, ctx, prod_dir, event=None):
         }
 
         return prod_json, product_staged_metadata, metrics
-    except (Exception, ):
+    except Exception:
         if event:
             event.set()
         logger.error("Publishing fialed on %s" % prod_dir)
```
mcayanan, line 881: typo? should be `failed`?

PR #165, opened 2023-05-15
rollback failed localization in backoff handler

DustinKLo:
change(s):
- ~rollback failed localization in `backoff` handler~
- wrap `try/except` in `osaka.main.get` in `download_file` add additional logging in backoff handler
  - rollback failed dataset localization in the `download_file` function


pymonger:
@DustinKLo: is this PR needed anymore?

PR #168, opened 2023-06-13
HC-471: Resolve time out issue when doing user rule evaluation

mkarim2017:
[Build Status](https://swot-pcm-ci.jpl.nasa.gov/job/force-branches/job/algermis-force_branch-E2E-test/6/)

### DIFF
--- a/hysds/user_rules_dataset.py
+++ b/hysds/user_rules_dataset.py
@@ -89,6 +89,13 @@ def update_query(_id, system_version, rule):
     return final_query
 
 
+@backoff.on_exception(
+    backoff.expo, Exception, max_tries=backoff_max_tries, max_value=backoff_max_value
+)
+def search_es(index, body):
+    return grq_es.es.search(index=index, body=body)
+
+
 def evaluate_user_rules_dataset(
     objectid, system_version, alias=DATASET_ALIAS, job_queue=JOBS_PROCESSED_QUEUE
 ):
@@ -133,7 +140,8 @@ def evaluate_user_rules_dataset(
 
         # check for matching rules
         try:
-            result = grq_es.es.search(index=alias, body=final_qs)
+            # result = grq_es.es.search(index=alias, body=final_qs)
+            result = search_es(index=alias, body=final_qs)
             if result["hits"]["total"]["value"] == 0:
                 logger.info("Rule '%s' didn't match for %s (%s)" % (rule_name, objectid, system_version))
                 continue
--- a/hysds/user_rules_dataset.py
+++ b/hysds/user_rules_dataset.py
@@ -93,7 +93,7 @@ def update_query(_id, system_version, rule):
     backoff.expo, Exception, max_tries=backoff_max_tries, max_value=backoff_max_value
 )
 def search_es(index, body):
-    return grq_es.es.search(index=index, body=body)
+    return grq_es.es.search(index=index, body=body, timeout=30)
 
 
 def evaluate_user_rules_dataset(
--- a/hysds/__init__.py
+++ b/hysds/__init__.py
@@ -3,6 +3,6 @@ from __future__ import print_function
 from __future__ import division
 from __future__ import absolute_import
 
-__version__ = "1.2.5"
+__version__ = "1.2.6"
 __url__ = "https://github.com/hysds/hysds"
 __description__ = "HySDS (Hybrid Cloud Science Data System)"


hysds/user_rules_dataset.py
```diff
@@ -89,6 +89,13 @@ def update_query(_id, system_version, rule):
     return final_query
 
 
+@backoff.on_exception(
+    backoff.expo, Exception, max_tries=backoff_max_tries, max_value=backoff_max_value
+)
+def search_es(index, body):
+    return grq_es.es.search(index=index, body=body)
```
DustinKLo, line 96: i'd recommend adding a `timeout` here as well
https://elasticsearch-py.readthedocs.io/en/v7.14.0/api.html#elasticsearch.Elasticsearch.search
> timeout – Explicit operation timeout

mcayanan:
@mkarim2017 @DustinKLo @pymonger 

Testing this change on the NISAR side: https://nisar-pcm-ci.jpl.nasa.gov/job/force-branches/job/mkarim-force_branch-E2E-test/87/

mcayanan (commented):
@mkarim2017 Could you please bump the version to 1.2.6?

https://github.com/hysds/hysds/blob/9d20afc9b3453d68d2af621d747568964f2289c8/hysds/__init__.py#L6

hysds/user_rules_dataset.py
mcayanan: @DustinKLo Should we set this timeout=30?

DustinKLo: 30 sounds good

mkarim2017:
> __version__ = "1.2.5"

Done

hysds/user_rules_dataset.py
DustinKLo, line 96: another suggestion is to return only the relevant fields needed, that can potentially return the data faster from ES
it looks like user rules doesnt need any of the fields returned so its safe to return an empty `_source`

https://elasticsearch-py.readthedocs.io/en/v7.14.0/api.html#elasticsearch.Elasticsearch.search
> `_source` – True or false to return the _source field or not, or a list of fields to return

DustinKLo (changes requested):

hysds/user_rules_dataset.py
mcayanan: @mkarim2017 what do you think?

mcayanan: Good catch. I think that makes sense in terms of user rule evaluation. @mkarim2017 @pymonger what do you think?

mkarim2017: > another suggestion is to return only the relevant fields needed it can potentially return the data faster from ES it looks like user rules doesnt need any of the fields returned so its safe to return an empty `_source`
> 
> https://elasticsearch-py.readthedocs.io/en/v7.14.0/api.html#elasticsearch.Elasticsearch.search
> 
> > `_source` – True or false to return the _source field or not, or a list of fields to return

@DustinKLo @mcayanan @gmanipon 
 I see it uses the whole doc in doc_res = result["hits"]["hits"][0]. Also if we want to use this function as generic like other functions can also can call it, we dont know what fields to return. May be it is better to have the query updated to only look for those fields that needed?

mcayanan:
> @mkarim2017 @DustinKLo @pymonger
> 
> Testing this change on the NISAR side: https://nisar-pcm-ci.jpl.nasa.gov/job/force-branches/job/mkarim-force_branch-E2E-test/87/
> 
> IN SWOT: https://swot-pcm-ci.jpl.nasa.gov/job/force-branches/job/mmkhan-force_branch-E2E-test/698/
> 
> [Build Status](https://swot-pcm-ci.jpl.nasa.gov/job/force-branches/job/mmkhan-force_branch-E2E-test/698/)

Looks like both SWOT and NISAR force branches passed with this PR

hysds/user_rules_dataset.py
mkarim2017: @mcayanan my mistake, I was looking at another ES call, not search. Updated the timeout to 30 sec.

pymonger: That's right. the `dataset_jpath` feature allows the HySDS-IO param configurations to path down to any field in the body of the triggering dataset.

pymonger (approved):

hysds/user_rules_dataset.py
pymonger: @DustinKLo: you make a great point though with optimizing our ES queries especially for trigger rule evaluation. The return of the entire doc can be a very heavy operation especially in some of the NISAR dataset cases. It would be best to track this as a separate ticket though so it stays on our radar.

DustinKLo (approved):

PR #174, opened 2023-06-26
HC-479: Dataset Ingest Bulk throws KeyError: "found" error | HC-478: WatchDog Scripts are incorrectly trying to make an ES update to the alias instead of the actual index

DustinKLo:
dev-e2e: 
- https://nisar-pcm-ci.jpl.nasa.gov/job/force-branches/job/dustinlo-force_branch-E2E-test/42
- https://nisar-pcm-ci.jpl.nasa.gov/job/force-branches/job/ci-force_branch-E2E-test/267/

# HC-479 - Dataset Ingest Bulk throws KeyError: "found" error
related ticket: https://hysds-core.atlassian.net/browse/HC-479

related PR(s):
- https://github.com/hysds/hysds/pull/154

Description:
- bug found in the `get_job_status` function in `hysds.utils` where its is trying to grab the job document by `_id` on `job_status-current`
`job_status-current` used to be the sole index, but since we moved to rolling indices (ex. `job_status-2023.06.23`) `job_status-current` is now an alias
- we can't get the document directly by alias so we need to use the `_search` API instead
- moving away from using `requests` and instead using the elasticsearch python API
- functions affected:
  - `get_job_status`
  - `query_dedup_job`
  - `fail_job`

# HC-478 - WatchDog Scripts are incorrectly trying to make an ES update to the alias instead of the actual index
related ticket: https://hysds-core.atlassian.net/browse/HC-478

Change(s):
- fixing `tag_timedout_workers` function to use the proper `_index` to update the document
- renaming `id` to `_id` because `id` is a builtin function for python

PR #175, opened 2023-07-02
hot-fix: Ignore 404 errors as we had before when making requests calls

mcayanan:
There appears to be an issue with the recent changes made in https://github.com/hysds/hysds/pull/174

Didn't notice this before, but it looks like in the querying of the dedup jobs, that PR updated the function to use mozart_es instead of the requests call:

image

However, it forgot to take into account gracefully ignoring 404 errors.


mcayanan:
Tested the changes and jobs are now successfully getting submitted again:

image

pymonger (commented):
Thanks for catching.

PR #179, opened 2023-08-15
HC-493: Updates to better handle jobs that fail due to connection timeouts but are never persisted back up to the job status

mcayanan:
This PR updates the system to better handle jobs where connection timeouts occur, but never get caught by the system. Therefore, at the operator level, the jobs appear "stuck" at job-started or job-queued.

Here are the changes:

**Orchestrator**
Add backoff to the run_job call when submitting a job to be queued. We also reversed the order such that we make the call first, then set the job status to `job-queued` rather than the other way around as it was in the previous implementation. Lastly, if the exponential backoff still fails after it hits its max retries, then we added the capability to set the job status to `job-failed`, followed by sending it off to user rule job evaluation.

**Process Events**
Additionally look for ConnectionError task failures. When these kinds of failures occur, then we make sure to fail the job.

**Process Events Tasks**
When failing the job, if we cannot find the job record in ES, apply exponential backoff to give time for the job record to be searchable. Once we find the job record, only set it to job failed if the job has not yet completed.

### DIFF
--- a/hysds/orchestrator.py
+++ b/hysds/orchestrator.py
@@ -424,32 +424,59 @@ def submit_job(j):
             # generate celery task id
             job_json["task_id"] = uuid()
 
-            # log queued status
-            job_status_json = {
-                "uuid": job_json["task_id"],
-                "job_id": job_json["job_id"],
-                "payload_id": task_id,
-                "payload_hash": payload_hash,
-                "dedup": dedup,
-                "status": "job-queued",
-                "job": job_json,
-            }
-            log_job_status(job_status_json)
-
-            # submit job
-            res = run_job.apply_async(
-                (job_json,),
-                queue=queue,
-                time_limit=time_limit,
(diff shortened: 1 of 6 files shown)


hysds/orchestrator.py
```diff
@@ -424,32 +424,69 @@ def submit_job(j):
...
+                    "payload_hash": payload_hash,
+                    "dedup": dedup,
+                    "status": "job-failed",
+                    "job": job_json,
+                    "context": context,
+                    "error": error_info,
+                    "short_error": get_short_error(error_info),
+                    "traceback": traceback.format_exc(),
+                }
+                log_job_status(job_status_json)
```
mcayanan, line 463: @pymonger @DustinKLo Do we still want to append to the `results` list in the `job-failed` case here?

pymonger (approved):

PR #187, opened 2024-03-15
HC-520: Fix job worker watchdog to properly handle cases where hard time limit was reached

mcayanan:
This PR resolves several issues seen with job worker watchdogs:

- Fix the calculation of the duration. Needs to be `total_seconds()`
- Sort the task status results by latest. This is because we only look at the top most result when determining what to do.
- Upon a task_failed status, propagate the error and traceback so that this eventually gets back up to Figaro

After making the changes, this was tested on the NISAR cluster and was verified that the job was properly moved to job failed with the timedout tag:

image

### DIFF
--- a/scripts/watchdog_job_timeouts.py
+++ b/scripts/watchdog_job_timeouts.py
@@ -69,7 +69,7 @@ def tag_timedout_jobs(url, timeout):
 
         if status == "job-started":
             # get task info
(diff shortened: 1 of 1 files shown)


DustinKLo (approved):

pymonger (approved):
LGTM

PR #191, opened 2024-09-10
HC-540: Cache the function objects being returned by orchestrator.py get_function method

philipjyoon:
instead of having to exec import every time. This seems to improve CPU performance 3x.

Testing performed:

- I ran a single orchestrator celery worker on a Mozart machine in foreground and disabled all that are normally started by supervisord. I ran it using the same command that was in supervisord.conf.
- I queued 2000 jobs using the existing code without the caching change. It took about 42 seconds. Average CPU utilization was ~70% eyeballing using top over those 42 seconds.
- I stopped the process; change the code to add caching; restarted. Stopped and then reset Mozart to clear out mozart_es. Queued up the same 2000 jobs again. It took about 22 seconds with average CPU utilization of ~40%
- Restored supervisord.conf to start the orchestrator celery workers nominally and ran the OPERA hySDS cluster in forward mode for a few PGE jobs. The system performed nominally.

### DIFF
--- a/hysds/__init__.py
+++ b/hysds/__init__.py
@@ -3,6 +3,6 @@ from __future__ import print_function
 from __future__ import division
 from __future__ import absolute_import
 
-__version__ = "1.3.3"
+__version__ = "1.3.4"
 __url__ = "https://github.com/hysds/hysds"
 __description__ = "HySDS (Hybrid Cloud Science Data System)"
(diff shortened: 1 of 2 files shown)


pymonger (dismissed):
Nice optimization!

philipjyoon:
@pymonger @mcayanan  The PR title and the version have been updated.

pymonger (approved):
LGTM

PR #192, opened 2024-09-23
HC-522-update: Updates to support moving away from OPS user

mcayanan:
This PR makes the following updates to support the latest container changes in moving away from having the Verdi container default to the OPS user. In addition, these updates/bug fixes are due to the latest testing efforts that were done within the HECC environment:

- The default `PODMAN_CFG` in the celeryconfig.py no longer sets the `--userns=keep-id` flag or the setting of the `-u` flag option by default. Therefore, `root` is the default user when running podman now, which means the host user in the podman world due to our rootless podman mode set up.
- `VERDI_HOME` defaults to `/root` now since the Verdi container now moves away from defaulting to an OPS user. `/root` is also now where the HySDS Core code resides in now.
- celeryconfig.py passes in the `HOST_USER` environment variable now because we need that in order to properly set the PODMAN user socket file now. This is due to the default user in podman being root now, where UID=0, but we start up the socket server with a socket file pertaining to the actual UID of the host. 
- When updating the volume source mounts, check if the beginning string matches the `verdi_home` argument instead of a general `if verdi_home in k`
- Fixed the code to default a missing `environment` area in `PODMAN_CFG` to a list instead of a dictionary.
- When persisting environment variables in the podman run command, use the actual environment variable value instead 

Housekeeping Update:
- Version was downgraded to 1.3.3 since PR https://github.com/hysds/hysds/pull/189 set it to this value already and that was the first commit since V5.2.1 release of HySDS Core

### DIFF
--- a/configs/celery/celeryconfig.py.tmpl
+++ b/configs/celery/celeryconfig.py.tmpl
@@ -134,6 +134,10 @@ PODMAN_CFG = {
   ]
 }
 
+# Use the PODMAN_SOCK setting if the UID in the container is expected to be different
+# than outside the container
+# PODMAN_SOCK = /run/user/01234/podman/podman.sock
+
 DATASET_PROCESSED_QUEUE = "dataset_processed"
 USER_RULES_DATASET_QUEUE = "user_rules_dataset"
 ON_DEMAND_DATASET_QUEUE = "on_demand_dataset"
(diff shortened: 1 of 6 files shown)


pymonger (dismissed):
LGTM

PR #193, opened 2024-10-03
HC-546: Update logstash to move a failed job over to a new job_failed index

mcayanan:
This PR is in conjunction with the following other PRs:

- sdscli: https://github.com/sdskit/sdscli/pull/109
- lightweight jobs: https://github.com/hysds/lightweight-jobs/pull/34

This PR updates the logstash configuration such that when it sees a failed job, it will make sure to move it from `job_status` to `job_failed`. Note that we don't want to introduce any rollover since the intent is to preserve this failed jobs for further analysis and put the onus on the end user to purge this failed jobs eventually.

**Testing**
See comments in https://hysds-core.atlassian.net/browse/HC-546

### DIFF
--- a/configs/logstash/indexer.conf.mozart
+++ b/configs/logstash/indexer.conf.mozart
@@ -58,7 +58,6 @@ filter {
   if [resource] == "job" and [status] == "job-failed" {
     ruby {
       code => "
-         logger.info('Processing event for job-failed status: ' + event.get('payload_id').to_s);
          original_index = event.get('[job][job_info][index]');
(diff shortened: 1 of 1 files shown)


pymonger (approved):
Nice. Didn't realize logstash had a way of deleting documents at the same time it indexes one.

mcayanan:
> Nice. Didn't realize logstash had a way of deleting documents at the same time it indexes one.

Ya Hook suggested that this solution could be done through logstash and so had to look up in the docs on how to do this.

PR #195, opened 2024-10-28
HC-556: Fix to make the auto-retrying of jobs work properly again

mcayanan:
This PR fixes an issue to allow the auto-retrying of jobs, via the Figaro trigger rules, to work properly once again. Prior to queueing up the job for user rule evaluation, we should check to see if the status is job-failed and if so, set the index to job_failed when queuing up the job for user_rule_evaluation. This allows the user_rule_evaluation to work properly as it will now query for the job in the correct index rather than the specific job_status partitioned index at which it originated.

This PR is in conjunction with https://github.com/hysds/lightweight-jobs/pull/35

For testing, I ran a one-day reprocessing test on a cluster in SWOT and verified that all the auto-retries due to worker termination were retried successfully. Also verified that we did not have any lingering leftover revoked jobs:

image

**Housekeeping Update**

- Fix the pytest as it was failing.

### DIFF
--- a/configs/logstash/indexer.conf.mozart
+++ b/configs/logstash/indexer.conf.mozart
@@ -71,20 +71,20 @@ filter {
     }
   }
 
-  if [resource] == "job" {
-    ruby {
-      code => "
-       current_time = Time.now.strftime('%Y-%m-%d %H:%M:%S.%N')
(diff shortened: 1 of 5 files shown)


pymonger (approved):
LGTM

PR #196, opened 2025-01-30
Enable SSL/TLS for gunicorn for Mozart and GRQ API endpoints

stirlingalgermissen:
This change requires updates to /etc/httpd/conf.d/00_mozart-ssl.conf which are managed by the SA team. Update http->https for Mozart/GRQ referenced endpoints in that config file


mcayanan:
@stirlingalgermissen Would you mind creating an equivalent HySDS Core ticket? Nominally, we name the branch the core ticket number so that when we do releases, we can reference that core ticket in the release notes.

stirlingalgermissen:
Core ticket: https://hysds-core.atlassian.net/browse/HC-560

PR #201, opened 2025-05-21
HC-571: Make cache optionally writeable; provide mechanism to prevent cache eviction; use HYSDS_ROOT_WORK_DIR for cache base

mcayanan:
This PR makes the following improvements regarding HySDS' caching feature:

### Optionally Make Cache Writeable
- Optionally make the cache directory writeable from within a PGE job. Currently, it's hard-coded to read-only, but there are cases where we need to localize inputs from within a job itself, so being able to leverage the cache would be beneficial in those use-cases.
- To make this writeable, set the new `CACHE_READ_ONLY` setting in the celery config to `false`

### Optionally Prevent Cache Eviction
- Added a new setting in the celery config called `EVICT_CACHE` that will tell Verdi whether or not it can clean up the cache directory if needed. When a Verdi worker is set up to run multiple jobs at a time, one thread may see the need to delete the cache to free up more space. But the issue with that is that another thread may be using that cache, so it would be problematic to delete the cache while another thread needs it. This feature prevents that from happening.

### Use HYSDS_ROOT_WORK_DIR for determining the cache location
- In the `download_file` function, it currently keys off of `ROOT_WORK_DIR` in the celery config to determine the cache location. This is normally set to` /data/work`. However, in the supervisord, the job worker can be set up to point to another working directory like `/data/work_1`. This is set through an environment variable called `HYSDS_ROOT_WORK_DIR`. So, conceptually, we would assume that the cache directory would be written there instead. Therefore, the update is to key off of `HYSDS_ROOT_WORK_DIR` instead and if for some reason that is not set, fall back to using `ROOT_WORK_DIR`. This is to maintain backwards compatibility.
- As a result of this update, we also needed to update the podman configuration to persist `HYSDS_ROOT_WORK_DIR` to the PGE container.
- Also made an update to the `download_file` function to optionally be able to specify a different root working directory for the cache if needed. 

Updates were vetted within SWOT's automated test and it passed:

image


pymonger (approved):
LGTM. We need this in NISAR for the `p4` and `p5` instance types that will have up to 16 concurrent verdi workers on each node.

PR #221, opened 2026-07-20
HC-639, HC-638: prevent process_events and watchdog scripts from clobbering finalized job docs; make worker exceptions picklable

pymonger:
## Summary

- **Bug.** A failed job's authoritative `job-failed` document -- real error, exit code, `duration`, and the `products_staged` entry that points at the triage dataset -- gets silently overwritten by a *supervisory* process that rebuilt the document from a stale `job-started` snapshot it read straight from OpenSearch (OS). It is a read-side race against the worker's own in-flight write.
- **Impact** (measured on one NISAR OPS cluster, ~4 months): **1,356 damaged docs** (529 from `event_processors`, 827 from the watchdog); **none** kept their `duration`. The triage *products* were fine in GRQ/S3 -- only the pointer to them was destroyed, so failures render in Figaro with no triage link and no metrics.
- **Fix.** Before any supervisory write, consult the **synchronous redis status key** the worker already sets race-free (and that nothing read until now); give the watchdog a grace period timed from the failure event; and derive that redis key's TTL from the job's own `time_limit` so it can't expire mid-run.
- **Proven on a live cluster.** With the race forced deterministically, the buggy build clobbers and the fixed build stays healthy; a genuinely dead worker is still reaped within the specified bound; and the reviewer's 2-million-doc backlog concern was reproduced and answered. 50 new unit tests.
- **Ships with HC-638** (picklable exceptions) -- required, see the **Release coupling** section below.

## Two tickets, one PR

| Ticket | What it is |
|---|---|
| [HC-639] | Failed job docs lose their triage link and metrics when a supervisory writer overwrites them from a stale `job-started` source. **This is the fix.** |
| [HC-638] | `WorkerExecutionError` / `OrchestratorExecutionError` are unpicklable, so celery substitutes `UnpickleableExceptionWrapper` in the result, the event, and the traceback. Small change, but it **must ship in the same release** as HC-639 -- see the **Release coupling** section below. |

---

## The bug in one picture

A failed job has exactly **one** authoritative writer: the verdi worker -- the process that actually ran the job. It catches the failure, runs triage, and logs a complete `job-failed` document, which travels an **asynchronous** pipeline to OS (`redis list -> logstash -> OS`, ~1s).

The worker is not the only writer, though. Several **supervisory writers** run alongside it -- background daemons that never ran the job themselves, but monitor job state and write a status document *on the worker's behalf* when they decide a job has failed or gone missing. Two are in this hot path:

- **`event_processors.fail_job`** -- fires on a celery `task-failed` event, to record failures a genuinely lost worker never got to report;
- **`scripts/watchdog_job_timeouts.py`** -- periodically sweeps OpenSearch for jobs still marked running past their timeout, and reaps them.

(`scripts/offline_orphaned_jobs.py` is the same pattern, run manually.) Each reads job state **directly from OS** and writes a whole document back. If one reads before the worker's own document is searchable, it rebuilds the job from the stale `job-started` snapshot and overwrites everything the worker recorded.

```mermaid
sequenceDiagram
    participant W as verdi worker
    participant R as redis
    participant L as logstash
    participant OS as OpenSearch job_status
    participant P as process_events

    Note over W: PGE fails. Worker catches it,<br/>runs triage, builds complete<br/>job-failed doc with metrics
    W->>R: SETEX status key = job-failed (synchronous)
    W->>R: RPUSH good doc (asynchronous)
    W--)P: celery task-failed event
    Note over P: RC1 regex false-match routes this<br/>handled failure into the lost-worker lane
    P->>OS: read job doc by uuid
    OS-->>P: stale job-started _source
    R-->>L: drain list
    L->>OS: index good job-failed doc
    Note over P: RC3 rebuild whole doc<br/>from the stale _source
    P->>R: RPUSH bad doc
    R-->>L: drain list
    L->>OS: bad doc replaces good doc<br/>metrics and triage link gone
```

Two facts make this subtle:

1. **Only the read is contested.** Every writer enqueues through the same redis list, so the bad document always lands *after* the good one. Nothing races at write time -- what races is whether the supervisory OS *read* beats the good document becoming searchable. When it loses, the rewrite is total (it rebuilds the whole doc), so metrics that were never in the stale `_source` cannot survive.
2. **The race-free path already existed and nobody used it.** The same `log_job_status()` call that enqueues the OS document also does a synchronous `SETEX` of the status onto a redis *key*. That key is correct the instant the worker finalizes. Consulting it is the whole fix.

## Why it happens -- three root causes

None of the three is a recent regression -- they have been latent in the supervisory-write paths since early in the codebase, which is why the damage accrued slowly and went unnoticed.

**RC1 -- a regex false-match opens the racing lane.** `TASK_FAILED_RE` decides which task-failed events mean "the worker had no chance to report". With no word boundary, `SoftTimeLimitExceeded` (a failure the worker fully handles and triages) matches `TimeLimitExceeded` and is routed into `fail_job`, which then races the worker's own in-flight document.

| event `exception` string | should enter `fail_job` | old regex | new regex |
|---|---|---|---|
| `WorkerExecutionError('SoftTimeLimitExceeded()')` | no | **yes** | no |
| `SoftTimeLimitExceeded()` | no | **yes** | no |
| `TimeLimitExceeded(3600,)` | yes | yes | yes |
| `WorkerLostError('...signal 9 (SIGKILL).')` | yes | yes | yes |
| `ConnectionError('Error 111 connecting to redis')` | yes | yes | yes |

The pattern gains a `(?<![A-Za-z])` lookbehind and moves into `hysds.event_processors` so it is importable and unit-tested; `scripts/process_events.py` imports it. (HC-638 alone does **not** close this lane -- the fixed string still contains the substring `TimeLimitExceeded`.)

**RC2 -- the watchdog is always armed.** Its eligibility clock runs from the `job-started` document's `@timestamp` (job *start* time), so any job older than the threshold is permanently eligible and re-swept every interval -- with zero grace once a task-failed event appears.

```mermaid
flowchart LR
    A["job-started doc<br/>@timestamp = job START time"] -->|"older than threshold<br/>= eligible"| B["permanently eligible,<br/>re-swept every interval"]
    B --> C{"task-failed<br/>event exists?"}
    C -->|"yes: first sweep, zero grace"| D["rewrite from stale _source"]
    C -->|no| B
```

Fix: a grace period measured **from the task-failed event** (`-g/--grace-secs`, default 300), plus the RC3 guard.

**RC3 -- every supervisory write was an unguarded read-modify-write.** Five write sites rebuilt whole documents from a stale `_source` with no concurrency control: `event_processors.fail_job`; the watchdog's transition write; the watchdog's *tag-only* write (which republishes the entire stale doc -- `status: job-started` included -- just to append a `timedout` tag, silently reverting a finished job); and both writes in `scripts/offline_orphaned_jobs.py`.

## The fix -- a gate in front of every supervisory write

The fix adds two checks in front of every supervisory write. Two shorthands, used from here on and in the tables below:

- **guard** -- the new `hysds.log_utils.is_job_finalized(task_id)`. It reads the *synchronous* redis status key (the one the worker [`SETEX`s the instant it finalizes](https://github.com/hysds/hysds/blob/ca436500300e50f71b30cdcf525154344c1ba596/hysds/log_utils.py#L294-L298)) to answer one question: *has the worker already written a terminal status for this job?* Every supervisory write consults it immediately before writing (never at loop top).
- **grace** -- the watchdog's new `-g/--grace-secs` wait (default 300s), measured from the task-failed event. It gives the worker's own document time to travel the async pipeline before the watchdog even considers acting.

The watchdog applies **both** gates; the other supervisory writers apply **only the guard**. The flow below is the **watchdog's** -- the most complex case, since it is the only one with a grace period and a tag lane:

```mermaid
flowchart TD
    A["watchdog's periodic OpenSearch scan<br/>turns up a job past its timeout"] --> B{"job document<br/>well-formed?"}
    B -->|no| B1["skip it and keep sweeping<br/>(one bad doc no longer<br/>aborts the whole sweep)"]
    B -->|yes| C{"is there a task-failed<br/>event for this job?"}

    C -->|yes| D{"has the grace period elapsed<br/>since that failure event?"}
    D -->|"not yet"| D1["defer -- write nothing<br/>(any write re-stamps @timestamp<br/>and restarts the sweep clock)"]
    D -->|"yes"| E{"does redis say the worker<br/>already finished this job?"}
    E -->|"yes"| E1["decline -- the worker's doc wins<br/>(log DIVERGENCE + divergence_count)"]
    E -->|"no, or key missing<br/>(fail open)"| F["reap -- write job-failed"]

    C -->|no| G{"did the job run past<br/>its time_limit?"}
    G -->|"yes"| H{"does redis say the worker<br/>already finished this job?"}
    H -->|"yes"| H1["decline -- tagging would republish<br/>the whole stale document"]
    H -->|"no, or key missing"| I["append the timedout tag"]

    classDef added fill:#2da44e,stroke:#1a7f37,stroke-width:2px,color:#ffffff;
    class B,B1,D,D1,E,E1,H,H1 added;
```

*Green = **added** by this PR (the guard, the grace period, the parse-hardening). The new checks are inserted around unchanged pre-existing boxes (the scan, the task-failed-event routing, the `job-failed` transition, the time_limit check, and the `timedout` tag write), so nothing in the watchdog was modified in place.*

The **event-driven** writer is much simpler. When a `task-failed` event clears the RC1 regex, `process_events.py` calls `event_processors.fail_job`, which runs the guard alone -- the same `is_job_finalized` check: a terminal redis key -> decline, otherwise write `job-failed`. It has no grace gate and no tag lane, and needs neither: the regex ensures only genuine lost-worker events reach it, and a lost worker never wrote a terminal status, so there is nothing in flight to wait for. `scripts/offline_orphaned_jobs.py` likewise applies only the guard, before each of its writes.

```mermaid
flowchart TD
    A["celery task-failed event arrives"] --> B{"does the exception match<br/>TASK_FAILED_RE?<br/>(a genuine lost-worker signature)"}
    B -->|"no -- e.g. SoftTimeLimitExceeded,<br/>now correctly excluded"| B1["ignore -- the worker<br/>handles and logs this itself"]
    B -->|yes| C{"does redis say the worker<br/>already finished this job?"}
    C -->|"yes"| C1["decline -- the worker's doc wins<br/>(log, do not overwrite)"]
    C -->|"no, or key missing<br/>(fail open)"| D["reap -- write job-failed<br/>+ requeue rule evaluation"]

    classDef added fill:#2da44e,stroke:#1a7f37,stroke-width:2px,color:#ffffff;
    classDef modified fill:#0969da,stroke:#0a3069,stroke-width:2px,color:#ffffff;
    class C,C1 added;
    class B modified;
```

*Blue = **modified** -- the pre-existing RC1 regex gained a word boundary (B). Green = **added** -- `fail_job`'s guard and its decline (C, C1). The event trigger and the write are pre-existing.*

The guard **fails open**: a missing key or a redis error means "proceed", because the watchdog is the last backstop for genuinely lost workers -- absence of evidence must not disable it. One more hole is closed alongside: the redis key's `SETEX` TTL is now `max(HYSDS_JOB_STATUS_EXPIRES, time_limit + 2 x HARD_TIME_LIMIT_GAP)`, so for long job types (e.g. `time_limit` >= the 86400s default expiry) the key no longer expires mid-run and leaves the guard blind.

### Changes by file

**HC-638 -- picklable exceptions**
- `hysds/job_worker.py` (`WorkerExecutionError`) and `hysds/orchestrator.py` (`OrchestratorExecutionError`): `job_status` defaults to `None`. Celery rebuilds exceptions via `cls(*exc.args)`. HC-582 had moved `job_status` out of `args` to shrink oversized tracebacks (they were exceeding 700KB), but it stayed a required constructor parameter, so `cls(*args)` raised `TypeError` and celery substituted `UnpickleableExceptionWrapper`. Defaulting the parameter keeps HC-582's smaller `args` (traceback-size contract preserved and tested) while letting reconstruction succeed. Dead `job_status(self)` methods removed from both classes and from `JobDedupedError`.

**HC-639 -- close the races**
- `hysds/log_utils.py`: `TERMINAL_JOB_STATUSES` + `is_job_finalized()`; TTL derived from the job's `time_limit`.
- `hysds/event_processors.py`: bounded `TASK_FAILED_RE`; `is_job_finalized` early-return in `fail_job`.
- `scripts/process_events.py`: imports the shared regex.
- `scripts/watchdog_job_timeouts.py` -- the watchdog gets the grace period and the guard, plus two robustness fixes:
    - new `-g/--grace-secs` flag sets the grace period (default 300s);
    - the sweep's query now also pulls the task-failed event's `@timestamp`, which the grace check needs to know how long ago the failure happened;
    - the grace check and the guard both run immediately before each of the watchdog's two writes -- the one that transitions a job to `job-failed`, and the one that appends the `timedout` tag;
    - when redis says a job finished but OpenSearch still shows it running, the watchdog logs a `DIVERGENCE` warning and tallies these per sweep (`divergence_count`) instead of overwriting the job;
    - each candidate document is now parsed defensively, so a single malformed document is skipped instead of aborting the entire sweep.
- `scripts/offline_orphaned_jobs.py` -- the same guard added before both of its write sites. It runs *before* the `--dry_run` check, so a dry run reports the decision a real run would actually make (including "declined"), not a plan the guard would later veto.
- `hysds/job_worker.py` -- one-line fix in the dedup path for a redelivered job (under `ENABLE_JOB_LOCKING`): the redis status key is now written under `task_id` instead of `job_id`. Every reader, the guard included, looks the key up by `task_id`, so the old `job_id` key was one nobody ever read -- meaning the guard could never tell that these jobs had already finished.

<details>
<summary><b>Edge cases considered</b> (click)</summary>

*Reminder:* **guard** = the `is_job_finalized(task_id)` redis check; **grace** = the `--grace-secs` wait after the failure event (both defined in **The fix** section above).

| Edge case | Pre-fix | With this PR |
|---|---|---|
| Soft-limit failure, worker alive and finishing | RC1 mislabels it; `fail_job` races the in-flight doc | regex no longer matches; if the lane is reached anyway, the guard declines |
| Long job at the TTL boundary (`time_limit` >= `HYSDS_JOB_STATUS_EXPIRES`) | redis key expires mid-run -> guard would see "no key" and proceed | TTL now derived from `time_limit`; key outlives the job |
| Genuinely lost worker (SIGKILL, spot reclaim, hard limit) | reaped (correct) | still reaped: fail-open guard; grace delays reaping by at most grace + one interval |
| Finished job swept by the tag-only path | whole stale `_source` republished; terminal status reverts to `job-started` | guard immediately before the tag write |
| Grace-deferral bookkeeping | n/a | the deferral path writes nothing -- any `log_job_status()` would re-stamp `@timestamp` and reset the doc's own clock |
| Malformed / partial job doc in the sweep | one bad subscript aborts the whole sweep | per-doc parse hardened; doc skipped, sweep continues |
| redis terminal while OS stuck at `job-started` (dropped logstash write) | unguarded watchdog "repaired" it invisibly by rewriting | guard declines + logs `DIVERGENCE` + `divergence_count` (the detector; reconcile arm deferred, see below) |
| transient `queue_finished_job` failure inside `fail_job` | n/a | guard runs once, OUTSIDE the backoff-retried body -- a retried guard would read the key its own first attempt wrote and drop the requeue |
| `ConnectionError` event for a live, already-finalized worker | document rewritten | guard declines |

</details>

---

## Validation on a live cluster

Everything below ran on `nisar-gmanipon-1628` (post-RSO) on hysds **v6.4.0 = the unfixed build**, against a snapshot-restored, read-only copy of the OPS FWD corpus (`job_status-*` 4.9M docs, `job_failed` 73,837, `task_status-*` 6.45M). The fix is mozart-side, so each iteration was a file swap + `supervisorctl restart` -- seconds, no redeploy.

**How the race was made deterministic:** freeze the *reader*. `refresh_interval: -1` on the index the job names in `job_info.index` **and** on `job_failed` (logstash reroutes a `job-failed` doc into `job_failed`, so stalling only the dated index would leave the good doc visible and the clobber would never fire). Every iteration asserts the stall actually took, else it scores VOID -- never PASS.

### PRE/POST -- the headline result

Identical scenario, `stall_effective: true` in both, only the build differs:

| Run | Build | `duration` | `len(tags)` | `_version` | `products_staged` | Outcome |
|---|---|:--:|:--:|:--:|---|---|
| control (fast failure, no stall) | buggy | 222.03 | 1 | 2 | `[triaged_job-…]` | HEALTHY |
| **PRE** (`soft=120`, stalled) | buggy | **absent** | 3 | 8 | **`[]`** | **CLOBBERED** |
| **POST** (identical, stalled) | fixed | present | 1 | 2 | **`[triaged_job-…]`** | **HEALTHY** |

`short_error` also flipped from the celery-event elision `WorkerExecutionError.....ceeded()')` back to the worker's real `SoftTimeLimitExceeded()`, and the exit code (255) survived. The clobbered fingerprint (`_version` high, array `id`, `len(tags)` grown) is the same direction seen in production; it reads `_version 8` here because the stall was held across several 60s watchdog ticks.

### The watchdog, PRE vs POST (cleanest pair -- one log file)

Pre-fix the watchdog didn't clobber once -- it rewrote the doc on **every** 60s sweep, which is what drove that document to `_version 8`:

```
17:02:02  Set job ddd73b3f... to job-failed via log_job_status().
17:02:57  Set job ddd73b3f... to job-failed via log_job_status().
17:03:58 / 17:05:13 / 17:06:17   (five rewrites, one per tick)
```

Post-fix, the same situation **defers** (a positive log line, not a silent absence -- a guard that imported the wrong `get_job_status` would also produce "0 clobbers", so absence would prove nothing):

```
17:23:48  Job ea000018...: task-failed 32s ago, within 300s grace; deferring to the worker.
```

### AC5 -- the grace delays reaping, it does not strand

The most important regression risk: a guard *without* a grace would strand genuinely dead jobs forever, since the watchdog is the last backstop. A job whose worker was SIGKILLed mid-cleanup (task failed `17:23:17`):

```
17:23:48  task-failed  32s ago, within 300s grace; deferring to the worker.
17:25:02 / 17:25:48 / 17:26:35 / 17:27:46   (age counting up)
17:28:34  Set job ea000018... to job-failed via log_job_status().   <-- REAPED
```

Reaped **317s** after the failure = grace (300s) + one poll interval (60s) -- exactly the specified bound. Bounded and observable.

### Split arms -- neither fix is masking the other

With the regex in, `fail_job` never runs on this path, so a clean POST says nothing about the guard. Each arm ran with the *other* fix disabled, each with a positive assertion:

| Arm | Config | Outcome | Positive assertion |
|---|---|---|---|
| guard-only | regex reverted | HEALTHY | `fail_job` ran, then declined |
| regex-only | guard forced False | HEALTHY | `fail_job` ran **0** times |

`fail_job` was invoked exactly once in the whole campaign -- the guard-only decline:

```
fail_job - 879bf66c-...: worker already finalized this job; not overwriting.
exc=WorkerExecutionError('SoftTimeLimitExceeded()'), short_error=TimeLimitExceeded
```

That one line carries three facts: `short_error=TimeLimitExceeded` is **RC1 caught in the act** (`match.group(1)` pulling the token out of the soft-limit string); `fail_job` genuinely executed (guard on the live path); and it declined. It also **excludes the import trap**: the stall was active, so OS read `job-started` at that instant -- a guard wired to `hysds.utils.get_job_status` (the OS reader) would have clobbered. It declined, so it is provably reading the synchronous redis key under production conditions.

### Reviewer concern: does a timer survive a deep backlog?

> *"I'm concerned about a timer based solution here. What if there are a million docs in the redis-list -> logstash -> ES pipeline queue? We have hit 2 million on OPERA."* -- @stirlingalgermissen

Right concern, and a fixed timer alone would **not** survive it. It isn't alone: the grace timer is the *first* of two gates; the load-bearing one is the redis guard, which reads a synchronous **key** that never enters the pipeline. I reproduced the exact 2M scenario -- `logstash_indexer` stopped (an unbounded backlog, worse than 2M), 2,000,000 entries pushed onto the `logstash` list, then a probe soft-limited so its terminal doc's `RPUSH` sits behind the 2M while OS still serves `job-started`:

```json
{ "list_backlog_depth": 2004230,
  "redis_key_get_job_status": "job-failed",
  "guard_is_job_finalized": true,
  "ES_doc_status": "job-started" }
```

It holds because `log_job_status()` writes the status **twice** -- `SETEX` on a key, `RPUSH` on the list. The backlog is the *list*; the guard does a `GET` on the *key*. Different data structures; a `GET` is O(1) and cannot be delayed by `LLEN`. A deeper queue makes OS *more* stale, i.e. makes the guard *more* necessary and no less effective. With the grace deliberately expired (aged task-failed injected 600s back), the watchdog declined every sweep and emitted `divergence_count=1`; when the 2M drained, the worker's doc landed intact (`_version 2`, `duration 124.9`, triage link present). The one residual -- the key *expiring* mid-run -- is closed by the `time_limit`-derived TTL above. (The `sleep()` change mentioned is HC-633, already on develop and inherited here.)

<details>
<summary><b>Backlog experiment -- full writeup</b> (click)</summary>

**Setup.** `nisar-gmanipon-1628`, hysds fix installed. `logstash_indexer` STOPPED (pipeline frozen = the limiting case of an unbounded backlog), then 2,000,000 dummy entries LPUSHed onto the `logstash` list to match the exact number cited. A probe job was soft-limited so the worker wrote its terminal `job-failed` document -- whose RPUSH now sits behind the 2M-deep backlog, undrained, while its `job-started` document (indexed before the freeze) is what OS still serves.

**The measurement.**
```json
{ "list_backlog_depth": 2004230,
  "resolved_task_id": "d6d0a919-18ee-44d1-8b93-3699d71844bc",
  "redis_key_get_job_status": "job-failed",
  "guard_is_job_finalized": true,
  "ES_doc_status": "job-started",
  "ES_products_staged": [] }
```

**Why the backlog cannot reach the guard.** `log_job_status()` (`hysds/log_utils.py`):
```python
r.setex(JOB_STATUS_KEY_TMPL % job["uuid"], ttl, job["status"])  # KEY  hysds-job-status-<task_id>
r.rpush(app.conf.REDIS_JOB_STATUS_KEY, msgpack.dumps(...))       # LIST "logstash"
```
The guard (`is_job_finalized` -> `get_job_status`) does `GET` on the **key**. The backlog is the **list**. A `GET` on the key is O(1) and cannot be delayed by `LLEN`.

**End-to-end, grace expired, under the 2M backlog.** An aged (600s) `task-failed` was indexed so the watchdog's transition path ran with the grace **already expired** -- the purest form of the objection:
```
23:22:04 WARNING watchdog_job_timeouts: Job 00202dcb-...: DIVERGENCE redis=job-failed es=job-started; declining rewrite.
23:22:04 WARNING watchdog_job_timeouts: divergence_count=1 docs with a terminal redis status but a stale non-terminal ES doc this sweep
23:23:06 WARNING watchdog_job_timeouts: Job 00202dcb-...: DIVERGENCE redis=job-failed es=job-started; declining rewrite.
```

**After the pipeline drains.** The 2M dummies were removed by exact value (`LREM`; the 6,652 real entries untouched) and `logstash_indexer` restarted. The worker's terminal document landed intact:
```json
{ "_index": "job_failed", "_version": 2, "status": "job-failed",
  "short_error": "SoftTimeLimitExceeded()", "duration": 124.945009, "exit_status": 255,
  "products_staged": ["triaged_job-probe-soft120-sleep300-...-task-d6d0a919-..."],
  "len_tags": 1, "id_type": "scalar", "traceback_has_wrapper": false }
```
Clean worker-written fingerprint, no clobber, after a 2M-deep backlog that outlasted the grace entirely.

</details>

### HC-638 -- verdi-side (worker) pickling

The mozart hot-patch can't reach `job_worker.py`/`orchestrator.py` (they run on the worker), so this was validated separately against the **deployed** hysds via `get_pickleable_exception` -- exactly celery's own call when recording a task failure.

| exception | deployed v6.4.0 (PRE) | after HC-638 (POST) |
|---|---|---|
| `WorkerExecutionError` | `pickle.loads` fails (`TypeError`) -> `UnpickleableExceptionWrapper` | `pickle.loads` OK -> `WorkerExecutionError` |
| `OrchestratorExecutionError` | fails -> `UnpickleableExceptionWrapper` | OK -> `OrchestratorExecutionError` |

So the task-failed event and traceback now carry the real class instead of the wrapper. (Run on factotum's writable `hysdsops` tree; the container worker runs celery as root and can't be hot-patched.)

### Unit tests

**50 new tests** across `test_exceptions.py`, `test_log_utils.py`, `test_event_processors.py`, `test_watchdog_job_timeouts.py`, `test_offline_orphaned_jobs.py`, each proven to fail on the parent commit where it targets changed behavior. Hardened against an adversarial mutation pass: the within-grace deferral is tested with a *tag-eligible* doc (kills a "log the deferral by tagging" mutant); the guard is a recording mock asserted *not* consulted during deferral (kills a loop-top-guard mutant); the backoff self-arm is covered end-to-end; and import-provenance pins tie `is_job_finalized` to `hysds.log_utils` (redis, not the identically-named OS reader in `hysds.utils`) and `process_events.py` to the shared regex object.

<details>
<summary><b>Test output + corpus census + raw run records</b> (click)</summary>

**Suite** (green; two pre-existing macOS-only `test_disk_usage` failures unrelated to this diff, verified failing on the parent of commit 1):
```
--- lock tests (isolated, per CI) ---    20 passed in 16.34s
--- all other tests ---                  2 failed, 124 passed, 1 skipped   (the 2 = pre-existing macOS test_disk_usage)
--- new test files ---                   50 passed in 0.91s
```

**Corpus census that sized the problem** (`b1`):
```
job-failed total                                        73837
job-failed missing duration                             5762
wrapper docs (damaged, HC-638 handle)                   1356
wrapper AND has duration (expect 0)                     0
-- writer cross-tab (perfect partition, 0 crossover) --
wrapper + time_end (event_processors)                   529
wrapper + NO time_end (watchdog)                        827
CROSSOVER (expect 0 / 0)                                0 / 0
-- context: missing-duration outside the wrapper cohort (separate cause, not clobber) --
failed, no duration, NOT wrapper                        4406
-- fingerprint: damaged vs clean (n=60 each) --
damaged: _version=3 len(tags)=2 id=array
clean  : _version=2 len(tags)=1 id=scalar
```

**Per-run records** (`rc_results.jsonl`, 5 runs). `POST-rc1` is reported AMBIGUOUS, not dropped: its worker was replaced by the ASG mid-run so it never reached the scripted failure -- the harness refuses to count a run that didn't reproduce the conditions.
```json
{"label":"PRE-rc1","stall_effective":true,"final":{"index":"job_failed","version":8,"status":"job-failed","short_error":"WorkerExecutionError.....ceeded()')","has_duration":false,"exit_status":null,"products_staged":[],"len_tags":3,"id_array":true,"wrapper":true},"outcome":"CLOBBERED"}
{"label":"POST-rc1","stall_effective":true,"final":{"index":"job_status-2026.07.21","version":2,"status":"job-started","len_tags":1,"id_array":false,"wrapper":false},"outcome":"AMBIGUOUS"}
{"label":"POST-rc1b","stall_effective":true,"final":{"index":"job_failed","version":2,"status":"job-failed","short_error":"SoftTimeLimitExceeded()","has_duration":true,"exit_status":255,"products_staged":["triaged_job-...-task-fd582741-..."],"len_tags":1,"id_array":false,"wrapper":false},"outcome":"HEALTHY"}
{"label":"GUARDONLY","stall_effective":true,"final":{"index":"job_failed","version":2,"status":"job-failed","short_error":"SoftTimeLimitExceeded()","has_duration":true,"exit_status":255,"products_staged":["triaged_job-...-task-879bf66c-..."],"len_tags":1,"wrapper":false},"outcome":"HEALTHY"}
{"label":"REGEXONLY","stall_effective":true,"final":{"index":"job_failed","version":2,"status":"job-failed","short_error":"SoftTimeLimitExceeded()","has_duration":true,"exit_status":255,"products_staged":["triaged_job-...-task-8fe27bbc-..."],"len_tags":1,"wrapper":false},"outcome":"HEALTHY"}
```

**Watchdog decision log** (`p4_watchdog_decisions.log`) -- pre-fix clobbers, post-fix grace deferrals, the bounded reap:
```
17:02:02 .. 17:06:17   Set job ddd73b3f... to job-failed via log_job_status().   (x5, pre-fix clobbers)
17:23:48 .. 17:27:46   Job ea000018...: task-failed <n>s ago, within 300s grace; deferring to the worker.   (x5)
17:28:34               Set job ea000018... to job-failed via log_job_status().   (reaped at +317s)
```

</details>

---

## Release coupling

**HC-638 and HC-639 must ship in the same release.** HC-638 removes the `UnpickleableExceptionWrapper` string -- which is the only search handle operators use to find already-damaged documents. That handle exists only because the HC-582 change caused the wrapper substitution in the first place, so the same behavior that produced HC-638 is what gives HC-639's damage its one fingerprint. Shipping HC-638 first would erase that fingerprint while leaving the data loss in place. The CHANGELOG records this. Merged together, reverted together, tested as the one configuration that ships.

This PR cuts **hysds 3.3.1** (`hysds/__init__.py` + the `[3.3.1]` CHANGELOG section).

## Deferred: the auto-reconcile arm (decided, not pending)

This PR ships the divergence **detector** (the watchdog warning + `divergence_count`). It does **not** ship an automatic reconcile that rewrites a diverged doc, and that is a deliberate decision:

- On the restored OPS corpus the OS-side strand proxy (a job doc non-terminal while its `task_status` is terminal) returned **1 doc in 4.9M**. But that number is weak on purpose: it's an instantaneous *stock* taken with the unguarded watchdog's clobber-repair still warm, and the OS-only proxy is biased *toward* "no strand". The true population (redis-terminal + OS-non-terminal) can't be measured on a dev corpus with an empty redis -- which is exactly what the shipped **detector** now measures live.
- A lost `job-failed` write **self-heals**: once the redis key expires the guard fails open and the watchdog reaps normally -- a delay, not a correctness loss. A lost `job-completed` write does **not** self-heal (the watchdog only ever transitions to failed/offline), but the only component with a completed arm (`scripts/offline_orphaned_jobs.py`) is manual and in no supervisord template.
- **Revisit trigger:** a non-zero `divergence_count` in production, or any observed lost `job-completed`, makes the reconcile arm required.

## Still outstanding before merge

- [ ] HC-638's `job_worker.py`/`orchestrator.py` hunks are verdi-side and were validated via the deployed pickling path, not a full worker job run -- confirm on a factotum-queued (host-celery) job type if desired.
- [ ] Reviewers: two approvals, one explicitly signing off on AC5.

[HC-639]: https://hysds-core.atlassian.net/browse/HC-639
[HC-638]: https://hysds-core.atlassian.net/browse/HC-638

### DIFF
--- a/test/test_event_processors.py
+++ b/test/test_event_processors.py
@@ -1,0 +1,212 @@
+import importlib.util
+import pathlib
+import sys
+import unittest.mock as umock
+
+import pytest
+
+# hysds.celery reads configuration on import, so mock it before any hysds
+# import. event_processors also instantiates a mozart ES client at module
+# scope (get_mozart_es -> reads ~/.netrc-os), so hysds.es_util must be mocked
+# too. FORCE both (not setdefault): under CI's single-process collection an
+# earlier test file imports the real modules first, which would make setdefault
+# a no-op and let the netrc read escape. unsafe=True because user_rules_job
+# does `from hysds.es_util import assert_doc_settled` and a default MagicMock
+# raises AttributeError for attributes starting with "assert". Snapshot and
+# restore afterwards so this does not pollute later test files (test_user_rules,
+# test_triage) that need the real modules.
+# restore es_util and the two modules that bind it at import, so the mock does
+# not leak into later test files. celery stays mocked (test_job_worker does the
+# same). event_processors stays cached on purpose -- the regex-provenance test
+# below relies on `pe.TASK_FAILED_RE is ep.TASK_FAILED_RE`, which requires the
+# script to import this exact module object.
+_saved = {k: sys.modules.get(k) for k in
+          ("hysds.es_util", "hysds.user_rules_job", "hysds.task_worker")}
+sys.modules["hysds.celery"] = umock.MagicMock()
+sys.modules["hysds.es_util"] = umock.MagicMock(unsafe=True)
+
+import hysds.event_processors as ep  # noqa: E402
+
+for _k, _v in _saved.items():
+    if _v is not None:
+        sys.modules[_k] = _v
+    else:
+        sys.modules.pop(_k, None)
+
+# (exception string, should trigger the ES-overwrite path)
+CASES = [
+    ("WorkerExecutionError('SoftTimeLimitExceeded()')", False),
+    ("SoftTimeLimitExceeded()", False),
+    (
+        "UnpickleableExceptionWrapper: WorkerExecutionError('SoftTimeLimitExceeded()')",
+        False,
+    ),
+    ("TimeLimitExceeded(3600,)", True),
+    ("celery.exceptions.TimeLimitExceeded(3600,)", True),
+    ("WorkerLostError('Worker exited prematurely: signal 9 (SIGKILL).')", True),
+    ("ConnectionError('Error 111 connecting to redis')", True),
+]
+
+
+@pytest.mark.parametrize("exc,should_match", CASES)
+def test_task_failed_re_classification(exc, should_match):
+    """AC1: SoftTimeLimitExceeded must not false-match TimeLimitExceeded."""
+    assert bool(ep.TASK_FAILED_RE.search(exc)) is should_match
+
+
+def test_guard_reads_redis_not_es():
+    """Two functions named get_job_status exist with opposite staleness:
+    hysds.log_utils (redis, correct) and hysds.utils (ES, the lagging store
+    this guard exists to avoid). Pin the import provenance -- a wrong import
+    passes every behavioral test here and does nothing in production."""
+    import hysds.log_utils as lu
+
+    assert ep.is_job_finalized is lu.is_job_finalized
+
+
+def test_fail_job_skips_when_worker_finalized(monkeypatch):
+    """AC2: no ES read and no write when redis says the job is terminal."""
+    monkeypatch.setattr(ep, "is_job_finalized", lambda uuid: True)
+    mock_es = umock.MagicMock()
+    monkeypatch.setattr(ep, "mozart_es", mock_es)
+    mock_log = umock.MagicMock()
+    monkeypatch.setattr(ep, "log_job_status", mock_log)
+
+    ep.fail_job({"traceback": "tb"}, "uuid-1", "exc", "short")
+
+    mock_es.search.assert_not_called()
+    mock_log.assert_not_called()
+
+
+def test_fail_job_proceeds_when_worker_not_finalized(monkeypatch):
+    """AC5: a genuinely lost worker is still reaped."""
+    monkeypatch.setattr(ep, "is_job_finalized", lambda uuid: False)
+    mock_es = umock.MagicMock()
+    mock_es.search.return_value = {
+        "hits": {
+            "total": {"value": 1},
+            "hits": [
+                {
+                    "_index": "job_status-current",
+                    "_source": {
+                        "status": "job-started",
+                        "payload_id": "p1",
+                        "uuid": "uuid-1",
+                        "job": {"job_info": {}},
+                    },
+                }
+            ],
+        }
+    }
+    monkeypatch.setattr(ep, "mozart_es", mock_es)
+    mock_log = umock.MagicMock()
+    monkeypatch.setattr(ep, "log_job_status", mock_log)
+    mock_finished = umock.MagicMock()
+    monkeypatch.setattr(ep, "queue_finished_job", mock_finished)
+
+    ep.fail_job(
+        {"traceback": "tb"},
+        "uuid-1",
+        "WorkerLostError('Worker exited prematurely: signal 9 (SIGKILL).')",
+        "WorkerLostError",
+    )
+
+    mock_log.assert_called_once()
+    written = mock_log.call_args[0][0]
+    assert written["status"] == "job-failed"
+    assert written["short_error"] == "WorkerLostError"
+    assert written["job"]["job_info"]["time_end"].endswith("Z")
+    mock_finished.assert_called_once()
+
+
+def test_script_imports_the_package_regex():
+    """scripts/process_events.py must consume the shared regex, not a stale
+    local copy -- reverting only the script hunk brings the false-match back."""
+    sys.modules.setdefault("future", umock.MagicMock())
+    scripts = pathlib.Path(__file__).resolve().parents[1] / "scripts"
+    spec = importlib.util.spec_from_file_location(
+        "process_events", scripts / "process_events.py"
+    )
+    pe = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(pe)
+    assert pe.TASK_FAILED_RE is ep.TASK_FAILED_RE
+
+
+def test_retry_does_not_self_arm_guard(monkeypatch):
+    """A transient queue_finished_job failure must not let the backoff retry
+    read the redis key the first attempt wrote and silently drop the requeue."""
+    state = {"finalized": False, "queue_calls": 0}
+    monkeypatch.setattr(ep, "is_job_finalized", lambda uuid: state["finalized"])
+    mock_es = umock.MagicMock()
+
+    def fresh_stale_response(*args, **kwargs):
+        # each retry re-queries ES, which is still serving the stale doc
+        return {
+            "hits": {
+                "total": {"value": 1},
+                "hits": [
+                    {
+                        "_index": "job_status-current",
+                        "_source": {
+                            "status": "job-started",
+                            "payload_id": "p1",
+                            "uuid": "uuid-1",
+                            "job": {"job_info": {}},
+                        },
+                    }
+                ],
+            }
+        }
+
+    mock_es.search.side_effect = fresh_stale_response
+    monkeypatch.setattr(ep, "mozart_es", mock_es)
+
+    def log_side_effect(job_status):
+        # the synchronous setex the real write performs
+        state["finalized"] = True
+
+    monkeypatch.setattr(
+        ep, "log_job_status", umock.MagicMock(side_effect=log_side_effect)
+    )
+
+    def queue_side_effect(payload_id, index=None):
+        state["queue_calls"] += 1
+        if state["queue_calls"] == 1:
+            raise ConnectionError("transient hiccup")
+
+    monkeypatch.setattr(
+        ep, "queue_finished_job", umock.MagicMock(side_effect=queue_side_effect)
+    )
+
+    ep.fail_job(
+        {"traceback": "tb"},
+        "uuid-1",
+        "WorkerLostError('Worker exited prematurely: signal 9 (SIGKILL).')",
+        "WorkerLostError",
+    )
+
+    assert state["queue_calls"] == 2
+
+
+def test_fail_job_leaves_terminal_es_doc_alone(monkeypatch):
+    """Second line of defence: the pre-existing ES-status else-branch."""
+    monkeypatch.setattr(ep, "is_job_finalized", lambda uuid: False)
+    mock_es = umock.MagicMock()
+    mock_es.search.return_value = {
+        "hits": {
+            "total": {"value": 1},
+            "hits": [
+                {
+                    "_index": "job_status-current",
+                    "_source": {"status": "job-failed", "payload_id": "p1"},
+                }
+            ],
+        }
+    }
+    monkeypatch.setattr(ep, "mozart_es", mock_es)
+    mock_log = umock.MagicMock()
+    monkeypatch.setattr(ep, "log_job_status", mock_log)
+
+    ep.fail_job({"traceback": "tb"}, "uuid-1", "exc", "short")
+
+    mock_log.assert_not_called()
--- a/test/test_event_processors.py
+++ b/test/test_event_processors.py
@@ -65,17 +65,23 @@ def test_guard_reads_redis_not_es():
 
 
 def test_fail_job_skips_when_worker_finalized(monkeypatch):
-    """AC2: no ES read and no write when redis says the job is terminal."""
+    """AC2: when redis says the job is terminal the guard declines -- no ES
+    read, no write, and no rule requeue. Skipping queue_finished_job here is
+    correct: the worker's own job-failed path already triggers rule
+    evaluation, so the supervisory path must not re-fire it."""
     monkeypatch.setattr(ep, "is_job_finalized", lambda uuid: True)
     mock_es = umock.MagicMock()
     monkeypatch.setattr(ep, "mozart_es", mock_es)
     mock_log = umock.MagicMock()
     monkeypatch.setattr(ep, "log_job_status", mock_log)
+    mock_finished = umock.MagicMock()
+    monkeypatch.setattr(ep, "queue_finished_job", mock_finished)
 
     ep.fail_job({"traceback": "tb"}, "uuid-1", "exc", "short")
 
     mock_es.search.assert_not_called()
     mock_log.assert_not_called()
+    mock_finished.assert_not_called()
 
 
 def test_fail_job_proceeds_when_worker_not_finalized(monkeypatch):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+## [3.3.1] - 2026-07-23
+
 ### Added
 - Python 3.12 compatibility updates
 - CHANGELOG.md file to track changes
@@ -17,6 +19,25 @@ All notable changes to this project will be documented in this file.
 - Updated test cases to work with Python 3.12
 
 ### Fixed
+- Failed job documents no longer lose their triage link (`products_staged`),
+  exit code, and duration to a stale overwrite by `event_processors.fail_job`
+  or the timeout watchdog (HC-639). Supervisory writers now consult the
+  synchronous redis status key (`is_job_finalized()`) immediately before
+  writing; the watchdog additionally waits a `-g/--grace-secs` grace period
+  (default 300s) measured from the task-failed event; `TASK_FAILED_RE` no
+  longer false-matches `SoftTimeLimitExceeded` as `TimeLimitExceeded`; and the
+  redis status-key TTL is derived from the job's own `time_limit`. The
+  watchdog also emits a per-sweep `divergence_count` WARNING (terminal redis
+  status with a stale non-terminal ES doc) as the health signal for lost
+  terminal writes. Behavior notes: `job-offline` counts as terminal for the
+  guard (timedout tagging declines on offline jobs whose redis key
+  survives), and a `ConnectionError` task-failed event for a worker that
+  already finalized no longer rewrites the job document.
+- `WorkerExecutionError`/`OrchestratorExecutionError` are picklable again, so
+  celery no longer reports `UnpickleableExceptionWrapper` in the result
+  backend and stored tracebacks (HC-638). MUST ship in the same release as
+  the HC-639 fix above: this change removes the traceback string operators
+  use to find documents damaged by HC-639.
 - Fixed `test_import` in `test_version.py` to use proper assertions
 - Resolved timezone-related `TypeError` exceptions in datetime operations
 
--- a/hysds/__init__.py
+++ b/hysds/__init__.py
@@ -1,3 +1,3 @@
-__version__ = "3.3.0"
+__version__ = "3.3.1"
 __url__ = "https://github.com/hysds/hysds"
 __description__ = "HySDS (Hybrid Cloud Science Data System)"
--- a/hysds/event_processors.py
+++ b/hysds/event_processors.py
@@ -61,10 +61,7 @@ def _fail_job(event, uuid, exc, short_error):
     query = {"query": {"bool": {"must": [{"term": {"uuid": uuid}}]}}}
 
     result = mozart_es.search(index="job_status-current", body=query)
-    # TODO: Remove this after debugging
-    logger.info(f"job status from fail_job: {json.dumps(result, indent=2)}")
     total = result["hits"]["total"]["value"]
-    logger.info(f"total results back from fail_job function: {total}")
     if total == 0:
         msg = f"Failed to query for task UUID {uuid}"
         logger.error(msg)
--- a/hysds/event_processors.py
+++ b/hysds/event_processors.py
@@ -3,6 +3,7 @@ from future import standard_library
 standard_library.install_aliases()
 
 import json
+import re
 import socket
 import traceback
 
@@ -17,6 +18,7 @@ from hysds.log_utils import (
     backoff_max_tries,
     backoff_max_value,
     get_val_via_socket,
+    is_job_finalized,
     log_job_status,
     logger,
 )
@@ -26,10 +28,35 @@ from hysds.user_rules_job import queue_finished_job
 
 mozart_es = get_mozart_es()
 
+# task-failed patterns meaning "the worker had no chance to send an update".
+# A soft time limit is not one: the worker catches it, triages, and logs its
+# own job-failed doc. The lookbehind keeps SoftTimeLimitExceeded from
+# matching TimeLimitExceeded.
+TASK_FAILED_RE = re.compile(
+    r"(?<![A-Za-z])(WorkerLostError|TimeLimitExceeded|ConnectionError)"
+)
 
-@backoff.on_exception(backoff.expo, Exception, max_tries=5, max_value=10)
+
 def fail_job(event, uuid, exc, short_error):
     """Set job status to job-failed."""
+
+    # log_job_status() is async on the ES side, so the ES read below can
+    # trail the worker's own terminal write; the redis key is authoritative
+    # (same re-check pattern as offline_jobs below). The guard sits OUTSIDE
+    # the retried body: _fail_job's own write sets the key to job-failed, so
+    # a retried guard would read its own write and drop the requeue.
+    if is_job_finalized(uuid):
+        logger.info(
+            f"fail_job - {uuid}: worker already finalized this job; "
+            f"not overwriting. exc={exc}, short_error={short_error}"
+        )
+        return
+    _fail_job(event, uuid, exc, short_error)
+
+
+@backoff.on_exception(backoff.expo, Exception, max_tries=5, max_value=10)
+def _fail_job(event, uuid, exc, short_error):
+    """Rewrite the job doc as job-failed and requeue rule evaluation."""
 
     query = {"query": {"bool": {"must": [{"term": {"uuid": uuid}}]}}}
 
--- a/hysds/job_worker.py
+++ b/hysds/job_worker.py
@@ -574,7 +574,9 @@ def run_job(job, queue_when_finished=True):
                 message = f"Redelivered job for payload {payload_id} - lock held by {lock_holder_info}. Deduping this job."
                 logger.warning(message)
                 job_status_json = {
-                    "uuid": job["job_id"],
+                    # uuid = celery task id (as at every sibling call site);
+                    # job_id would key a redis entry nobody reads
+                    "uuid": job["task_id"],
                     "job_id": job["job_id"],
                     "payload_id": payload_id,
                     "payload_hash": payload_hash,
--- a/hysds/job_worker.py
+++ b/hysds/job_worker.py
@@ -360,13 +360,12 @@ def redelivered_job_dup(job):
 
 
 class WorkerExecutionError(Exception):
-    def __init__(self, message, job_status):
+    # celery rebuilds exceptions via cls(*exc.args); params outside args need
+    # defaults or the class degrades to UnpickleableExceptionWrapper
+    def __init__(self, message, job_status=None):
         self.message = message
         self.job_status = job_status
         super().__init__(message)
-
-    def job_status(self):
-        return self.job_status
 
 
 class JobDedupedError(Exception):
@@ -374,9 +373,6 @@ class JobDedupedError(Exception):
         self.message = message
         self.job_status = "job-deduped"
         super().__init__(message)
-
-    def job_status(self):
-        return self.job_status
 
 
 def shutdown_worker(celery_hostname):
--- a/hysds/log_utils.py
+++ b/hysds/log_utils.py
@@ -226,6 +226,35 @@ def get_job_status(task_id):
     return res.decode() if hasattr(res, "decode") else res
 
 
+# statuses after which an authoritative terminal doc has been logged.
+# job-offline is included: a job never leaves it on its own, so any
+# supervisory rewrite over it is destructive.
+TERMINAL_JOB_STATUSES = frozenset(
+    {"job-completed", "job-failed", "job-deduped", "job-offline", "job-revoked"}
+)
+
+
+def is_job_finalized(task_id):
+    """Return True if a terminal status was already logged for this task.
+
+    Reads the synchronous redis key set by log_job_status(), authoritative
+    while the ES doc is still in flight (redis list -> logstash -> ES).
+    Supervisory writers must check this before rewriting a job doc. Fails
+    open (False) on a missing key or redis error.
+    """
+    try:
+        status = get_job_status(task_id)
+    except Exception as e:
+        logger.warning(f"is_job_finalized({task_id}): redis check failed: {e}")
+        return False
+    if status is None:
+        logger.warning(
+            f"is_job_finalized({task_id}): no redis job-status key (missing or expired)"
+        )
+        return False
+    return status in TERMINAL_JOB_STATUSES
+
+
 @backoff.on_exception(
     backoff.expo, RedisError, max_tries=backoff_max_tries, max_value=backoff_max_value
 )
@@ -255,9 +284,16 @@ def log_job_status(job):
     # send update to redis
     r = StrictRedis(connection_pool=JOB_STATUS_POOL,
                     ssl_ciphers=app.conf.get("broker_use_ssl", {}).get("ciphers"))
+    # derive the key TTL from the job's own time_limit: a fixed TTL expires
+    # mid-run for any job that outlives it, silently disabling
+    # is_job_finalized() for exactly the jobs most likely to soft-limit
+    ttl = app.conf.HYSDS_JOB_STATUS_EXPIRES
+    time_limit = job.get("job", {}).get("job_info", {}).get("time_limit")
+    if isinstance(time_limit, int):
+        ttl = max(ttl, time_limit + 2 * hard_time_limit_gap())
     r.setex(
         JOB_STATUS_KEY_TMPL % job["uuid"],
-        app.conf.HYSDS_JOB_STATUS_EXPIRES,
+        ttl,
         job["status"],
     )
     r.rpush(app.conf.REDIS_JOB_STATUS_KEY, msgpack.dumps(job_for_indexing))  # for ES
--- a/hysds/log_utils.py
+++ b/hysds/log_utils.py
@@ -289,8 +289,11 @@ def log_job_status(job):
     # is_job_finalized() for exactly the jobs most likely to soft-limit
     ttl = app.conf.HYSDS_JOB_STATUS_EXPIRES
     time_limit = job.get("job", {}).get("job_info", {}).get("time_limit")
-    if isinstance(time_limit, int):
-        ttl = max(ttl, time_limit + 2 * hard_time_limit_gap())
+    # accept int or float, but not bool (an int subclass); mirrors the
+    # watchdog's check so a float time_limit never leaves the TTL unextended.
+    # int() the value so the setex TTL stays an integer number of seconds.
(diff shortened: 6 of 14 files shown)


test/test_event_processors.py
```diff
@@ -0,0 +1,125 @@
...
+# `from hysds.es_util import assert_doc_settled`, and a default MagicMock
+# raises AttributeError for any attribute starting with "assert".
+sys.modules.setdefault("hysds.celery", umock.MagicMock())
+sys.modules.setdefault("hysds.es_util", umock.MagicMock(unsafe=True))
+
+import hysds.event_processors as ep  # noqa: E402
+
+# (exception string, should trigger the ES-overwrite path) -- the test oracle
+# from HC-639 section 4
+CASES = [
```
stirlingalgermissen, line 37: This doesn't seem ideal - can we not map these to the actual messages in case they change? I see a hardcoded timelimitexceeded number that may or not be the actual value used in the future?

stirlingalgermissen:
I'm concerned about a timer based solution here. What if there are a million docs in the redis-list -> logstash -> ES pipeline queue? We have hit 2 million on OPERA, though that should be less likely happen with the recent sleep() change

pymonger:
Good objection, and it pushed on the right spot -- a fixed timer alone would not survive a deep backlog. It isn't alone: the grace timer is the *first* of two gates. The load-bearing one is the redis guard, which reads a synchronous **key** (`GET hysds-job-status-<task_id>`) that never enters the `logstash` list -- so it's O(1) and holds at any queue depth. A deeper queue only makes ES *more* stale, i.e. makes the guard *more* necessary and no less effective.

I reproduced your exact scenario -- `logstash_indexer` stopped and 2,000,000 entries pushed onto the list -- and the guard read `job-failed` while ES still served `job-started`; with the grace deliberately expired the watchdog declined every sweep and emitted `divergence_count=1`, and the worker's doc landed intact once the backlog drained. The full reproduction is now in the PR description under **Reviewer concern: does a timer survive a deep backlog?**

The one residual -- the redis key *expiring* mid-run -- is closed separately here by deriving the key TTL from the job's `time_limit`. (And the `sleep()` change you mentioned is HC-633, already on develop, so this branch inherits it.)

mcayanan (commented):
## Review Summary

Reviewed the full diff at head `ca43650`. The fix is well-reasoned, well-tested (50 new tests), and the root-cause analysis is sound. The core mechanism — gating supervisory writes on the synchronous redis status key via `is_job_finalized()` plus a watchdog grace period — is the right approach and defends in depth (redis guard *and* the pre-existing ES-status check in `_fail_job`).

Below are concerns worth raising before merge, in priority order.

## Priority Concerns

### 1. ⚠️ Type-handling mismatch on `time_limit` (real bug risk)

The two halves of the fix disagree on what a valid `time_limit` is:

**In `log_utils.py` (lines 290-293):**
```python
ttl = app.conf.HYSDS_JOB_STATUS_EXPIRES
time_limit = job.get("job", {}).get("job_info", {}).get("time_limit")
if isinstance(time_limit, int):
    ttl = max(ttl, time_limit + 2 * hard_time_limit_gap())
```

**In `watchdog_job_timeouts.py` (lines 62-66):**
```python
time_limit = src["job"]["job_info"]["time_limit"]
if isinstance(time_limit, bool) or not isinstance(
    time_limit, (int, float)
):
    raise TypeError(f"non-numeric time_limit: {time_limit!r}")
```

The watchdog accepts `(int, float)`, but the TTL derivation in `log_utils.log_job_status` only extends the TTL when `isinstance(time_limit, int)`. If any job carries a **float** `time_limit`, the redis key TTL is *not* extended → the key can expire mid-run → `is_job_finalized()` fails open → the exact clobber this PR prevents can recur for long-running jobs.

**Recommendation:** Use `isinstance(time_limit, (int, float)) and not isinstance(time_limit, bool)` in both places for consistency. (`bool` is an `int` subclass, so `True`/`False` currently sneak through the log_utils check too.)

### 2. `job["task_id"]` key access in the redelivered-dedup path

In `job_worker.py` line 577, this changes the dedup status doc's `uuid` from `job["job_id"]` to `job["task_id"]`:
```python
job_status_json = {
    # uuid = celery task id (as at every sibling call site);
    # job_id would key a redis entry nobody reads
    "uuid": job["task_id"],
```

The change is correct in intent (so `is_job_finalized` can find the key), but it's a direct `[]` access. Worth confirming `task_id` is *always* present on `job` at this point in the redelivery branch — a missing key now raises `KeyError` where the old code did not.

### 3. Fail-open leaves residual clobber risk (acknowledged, but flag for ops)

`is_job_finalized()` returns `False` on missing key or any redis error. Under a redis outage, or after key expiry, supervisory writers proceed and can still overwrite. The PR acknowledges this ("self-heals" for `job-failed`), and the TTL-from-`time_limit` change mitigates expiry — but concern #1 partially undermines that mitigation for float time_limits.

Net: the guarantee is "no clobber *while redis is healthy and the key lives*," not absolute.

### 4. Increased reaping latency for genuinely-lost workers

A SIGKILLed worker that still emitted a `task-failed` event now waits `grace_secs` (default 300s) *plus* up to the sweep interval (default up to ~3600s) before being reaped. That's a deliberate trade, but it's a behavior change operators should know — failed jobs stay visibly "running" longer. Confirm this is acceptable for downstream PCM forward-processing latency SLAs.

### 5. `fail_job` early-return skips `queue_finished_job` (rule requeue)

When the guard trips in `event_processors.py` line 48, the supervisory path returns without calling `queue_finished_job` (user-rule re-evaluation). This is correct *only if* the worker's own `job-failed` path already triggers rule evaluation. That's the expected design, but it's the one behavioral path not directly asserted by the new tests — worth an explicit confirmation.

## Minor / Nits

- **Deployment plumbing:** the new `-g/--grace-secs` relies on its in-code default (300) unless supervisord templates pass it. Verify the watchdog invocation in the puppet/verdi/mozart supervisord templates doesn't need updating and that 300s is the intended prod default.
- **Config dependencies:** `app.conf.HYSDS_JOB_STATUS_EXPIRES` and `HARD_TIME_LIMIT_GAP` are now on a hot path; confirm both are defined in the shipped `celeryconfig` template (missing attr → `AttributeError` in `log_job_status`).
- **Leftover debug logging:** `_fail_job` still carries `# TODO: Remove this after debugging` that logs the full ES result at INFO — noisy on every supervisory failure; good opportunity to drop it while touching this function.
- **Double redis read:** the watchdog divergence branch calls `get_job_status(task_id)` again right after `is_job_finalized(task_id)` already read it. Harmless, marginally wasteful.
- **Test fragility:** the new tests do heavy module-scope `sys.modules` mocking and import-by-path; the comments show awareness of CI cross-test pollution, but this style is brittle to collection ordering. Not blocking.

## Verdict

No blocking correctness defect in the happy path, and it's a genuine, well-validated data-loss fix. **Concern #1 (float `time_limit` TTL gap)** is the one I'd want addressed before merge, since it can silently reopen the race for exactly the long-running jobs the TTL change targets. #2 and #5 are quick confirmations.

pymonger:
Thanks for the thorough review, @mcayanan. All points addressed below; the code changes are in follow-up commit `0095e83` (on top of the `ca43650` you reviewed, so the delta is isolated).

## Priority concerns

**1. ⚠️ Float `time_limit` TTL gap — fixed.** Good catch, this is the real one. `log_utils.log_job_status` now uses the same predicate as the watchdog:
```python
if isinstance(time_limit, (int, float)) and not isinstance(time_limit, bool):
    ttl = max(ttl, int(time_limit) + 2 * hard_time_limit_gap())
```
Two things beyond your suggestion: `not isinstance(bool)` (as you noted, `bool` is an `int` subclass), and `int(time_limit)` — a raw float would flow into `r.setex(...)`, and Redis `SETEX` requires integer seconds, so accepting a float without coercion would just move the failure. Added a regression test (`test_status_key_ttl_extends_for_float_time_limit`) that asserts a float extends the TTL and the result stays `int`; it fails against the parent commit.

**2. `job["task_id"]` direct access — confirmed always present, no change.** `job["task_id"]` is the unguarded pattern at **24 sites** across `job_worker.py` (every status-doc construction). The line this PR changed previously read `job["job_id"]` — also a direct `[]` access — so the KeyError surface is unchanged; if `task_id` could be missing here, 23 other sibling writes would already `KeyError`. Left as a direct access to match the file.

**3. Fail-open residual risk — acknowledged, and #1 tightens it.** With the float fix, the TTL-from-`time_limit` mitigation now covers float limits too, so key-expiry no longer reopens the window for exactly the long-running jobs you flagged. The remaining fail-open (redis outage) is the intended backstop — the watchdog must still reap genuinely lost workers when redis is unreachable — and the new `divergence_count` metric surfaces the one case where fail-open could still clobber (redis terminal + ES stale). Agreed the guarantee is "no clobber while redis is healthy and the key lives," not absolute.

**4. Reaping latency — deliberate; quantified.** Worst case is `grace` + one sweep interval. The hysds default interval is 3600s (so ~3900s on defaults), but this is operator-tuned and the deployment this was validated on runs `-i 60`, i.e. ~360s. Importantly, the delay is before the failure is *recorded*, not before anything reprocesses — a genuinely dead job's work already stopped; we're only deferring when the doc flips to `job-failed` so the worker's own authoritative write can land first. No downstream reprocessing SLA is affected; the visible effect is a failed job showing "running" for up to grace+interval. Flagging for PCM confirmation all the same.

**5. `queue_finished_job` skip on guard-decline — confirmed and now asserted.** Correct: the guard declines only when the worker already finalized, and the worker's own `job-failed` path calls `queue_finished_job`, so re-firing it from the supervisory path would double-requeue rule evaluation. Strengthened `test_fail_job_skips_when_worker_finalized` to assert `queue_finished_job` is **not** called on the decline path, with a docstring documenting why.

## Nits

- **Leftover debug logging — removed.** Dropped the full-ES-result INFO dump (and its `# TODO`) in `_fail_job`.
- **`-g/--grace-secs` deployment plumbing:** the in-code default (300) is the intended prod default; the mozart supervisord template doesn't need to pass `-g`. NISAR runs the watchdog `-t 300 -i 60`; the template would gain `-g <n>` only if we later size grace above 300 (no live evidence for that yet).
- **Config deps:** `HYSDS_JOB_STATUS_EXPIRES` and `HARD_TIME_LIMIT_GAP` aren't new hot-path deps — `HYSDS_JOB_STATUS_EXPIRES` is already read at other `log_utils` sites (the task-worker key, the dedup key), and `HARD_TIME_LIMIT_GAP` backs the pre-existing `ensure_hard_time_limit_gap()`. Both ship in the standard `celeryconfig` template.
- **Double redis read:** kept intentionally. `is_job_finalized()` is the fail-open predicate used at every guard site (including `event_processors`); inlining `get_job_status()` + `TERMINAL_JOB_STATUSES` membership at the two watchdog sites to save the read would duplicate the fail-open `try/except` and diverge from that shared abstraction. The extra `GET` is O(1) and only on the rare divergence/decline path. Happy to inline if you'd rather.
- **Test fragility:** agreed. The collection-order pollution you're sensing already bit CI once (the `es_util` `sys.modules` stub no-oped when an earlier test imported the real module) and is fixed on this branch via force-mock-with-restore. The remaining module-scope mocking is brittle-by-nature for these framework modules but contained, and CI is green.

Net: #1 fixed with a regression test; #2 and #5 confirmed (#5 now asserted). Ready for another look.

mcayanan (approved):
LGTM. I re-ran the follow up through AI and it indicates that there are no more concerns

mcayanan:
## Re-assessment: All Concerns Addressed

Thanks for the detailed responses, @pymonger. I've reviewed the follow-up commit `0095e83` and your explanations—all priority concerns are now resolved or properly justified.

### ✅ **Concerns Addressed**

**#1 - Float `time_limit` handling (CRITICAL) - FIXED**
- The type check now mirrors the watchdog: `isinstance(time_limit, (int, float)) and not isinstance(time_limit, bool)`
- Properly coerces to `int(time_limit)` for Redis `SETEX` compliance
- Test coverage confirms float handling and integer TTL result
- **This was the critical data-loss path—now closed.**

**#2 - `job["task_id"]` direct access - JUSTIFIED**
- Confirmed this is a file-wide convention (24 sites across `job_worker.py`)
- The PR changed `job["job_id"]` → `job["task_id"]`, both direct accesses
- KeyError surface area unchanged from baseline
- **Matching established patterns is correct; concern withdrawn.**

**#3 - Fail-open residual risk - MITIGATED**
- The float `time_limit` fix eliminates key-expiry risk for long-running jobs (the scenario I flagged)
- Remaining fail-open case (redis outage) is by design and observable via `divergence_count`
- Trade-off properly documented: "no clobber while redis is healthy and the key lives"
- **Acceptable design choice; concern resolved.**

**#4 - Reaping latency - QUANTIFIED**
- Worst-case: `grace + interval` (e.g., 360s on deployments with `-i 60`)
- Affects only when failure is *recorded*, not when work stops
- No downstream SLA impact; visible effect is delayed status flip
- **Design trade-off clearly documented.**

**#5 - `queue_finished_job` skip - CONFIRMED**
- Worker's own job-failed path already triggers rule evaluation
- Test now asserts supervisory path does NOT double-requeue
- **Correct behavior, now with regression coverage.**

### 📝 **Minor Items**
- Debug logging removed
- Deployment plumbing clarified (300s default appropriate)
- Double redis read justified (consistent with shared abstraction)

---

**Verdict: Ready to merge.** The critical fix (#1) closes the data-loss window, and all other concerns are either addressed or properly justified with technical rationale. Excellent work on the validation methodology and test coverage.

PR #223, opened 2026-09-01
HC-648: Stop late job-failed writes from resurrecting a deleted job doc, and reap the ones that get through

pymonger:
## Summary

When a failed job is retried, lightweight-jobs retry.py deletes its status doc before resubmitting. A job-failed write still in flight can land after that delete and re-create the doc, leaving a `job_failed` doc for an attempt that has already been superseded, sitting in the `job_status-current` alias next to the retried attempt's own doc. HC-640 fixes the delete side. This closes the write side and repairs the residue.

Resolves [HC-648](https://hysds-core.atlassian.net/browse/HC-648). Relates to [HC-640](https://hysds-core.atlassian.net/browse/HC-640), hysds/lightweight-jobs#45. hysds/sdscli#117 declares the two fields this relies on. Land order: sdscli#117, lightweight-jobs#45, then this.

## The three lanes this closes

**Duplicate terminal writes.** `run_job`'s failure tail logged the terminal status, queued its rules against `job_failed`, then called `fail_job()`, which logged the same doc again milliseconds later, on every failed job. `event_processors._fail_job` had the same defect across its backoff replays: `queue_finished_job` backs off only on `socket.error`, and celery raises kombu's `OperationalError` for an unreachable broker, which is not one, so a broker outage escaped into `_fail_job`'s own `Exception` backoff and re-ran `log_job_status` on every replay — up to five terminal writes for one `_id`, each with a paired logstash delete. The retry rule can fire about a second after the first write settles, so a later copy is still in the redis to logstash to OpenSearch pipeline when the retry's delete lands, and it re-creates the doc.

**Pipeline-late arrivals.** A message rpushed before the retry ran but indexed after the delete, during an indexer backlog. No writer-side guard can catch a message already in the pipe; only repair reaches it.

**Future or unknown writers.** The #221 guards are call-site guards, so a new writer or a regression re-opens the hole silently.

For the historical POP1 specimen, the `WorkerExecutionError.....` short-error shape and a traceback carrying celery's `trace_task` frame both point at the pre-#221 watchdog or event_processors lane. That stack predates #221's release and both lanes are already guarded on develop.

### One terminal write per failed job

`fail_job(job_status_json, jd_file, log_status=True)`, and the tail call site passes `log_status=False` because that doc was logged moments earlier and its rules already queued.

The handler needed the same treatment. The terminal doc is written inside the same `try`, so any failure after it — `queue_finished_job` exhausting its backoff, or SIGTERM during the publish on a verdi restart — reached a handler that rebuilt `job_status_json` and called `fail_job` at the default. Second write, same `_id`, deterministic for every failing job for the duration of the outage rather than a race. A `status_logged` flag now tracks whether the doc was handed to `log_job_status`; when it was, the handler keeps that doc, records the tail failure with `logger.error` and raises without writing again. That also stops the real `error`, `short_error` and `traceback` being overwritten by the tail exception.

`_fail_job` keeps a `written` flag across its backoff replays. Once the terminal doc is handed to `log_job_status`, a replay goes straight to the rule requeue and does not re-read the doc: by then the search can return our own job-failed write, and the already-terminal branch would have dropped the requeue. Backoff exhaustion emits a `fail_job_gave_up` event instead of only a log line.

The `run_job` call site is covered at the source level, with `ast`: no `fail_job` call after the terminal write may omit `log_status`. Reverting the tail call to its bare form restores the double write while every behavioural test stays green, because those exercise `fail_job` directly.

### A guard before every supervisory write

`job_supersession()` in `hysds/log_utils.py`, wired into `event_processors._fail_job` and `offline_jobs`, both of the watchdog's write guards, `task_revoked_handler`, the job-lock contention path and `offline_orphaned_jobs`. Same guard-immediately-before-the-write style as #221, taking the caller's client rather than building a second one.

It is one realtime `mget` over every index a job doc can live in: `job_failed` plus every daily from the caller's date to today, oldest first, capped at 62 with the caller's own index always included. Never a refresh-bound search, since searching for a doc moving between those indices is the defect class being guarded. A doc-ID read cannot be aimed at the alias — OpenSearch rejects a single-index op on a multi-index alias — so the members are enumerated. The home only moves forward: a retry rewrites `job_info.index` to the current daily, so an attempt started on D1 and retried on D3 has its live doc in D3 while the caller holds D1. A daily that was never created is a clean miss.

It returns four states:

- **SUPERSEDED** — a doc with a different uuid and a higher `job.retry_count`. A different uuid alone is not enough: an older attempt's leftover, an orphaned job_failed doc or an unswept job-revoked doc, also carries a uuid that is not ours, and declining to write because of one would drop a legitimate failure. A missing key counts as 0 on both sides, so a tie does not supersede.
- **OWNED** — nothing newer.
- **ABSENT** — no live doc in any home. Not ambiguous: it is the signature of retry.py having just deleted the doc while its replacement is still in the pipe. Writing then does not merely orphan — the doc carries the old attempt's `job_info.index`, so logstash's paired delete destroys the retried attempt's brand-new doc, and the reaper cannot repair that because no dated candidate is left.
- **UNKNOWN** — the probe could not ask. A transport fault, never read as deletion. During a rolling node restart a search still returns the hit while the doc-ID probes raise, and reading that as ABSENT would fail closed and lose every failure record in flight.

`_fail_job` retries its read on UNKNOWN and raises on ABSENT so its backoff re-reads. The mozart daemons write only on OWNED. The two verdi sites, `task_revoked_handler` and the lock path, write on OWNED and UNKNOWN, degrading to their previous unconditional write rather than dropping a terminal record when mozart's OpenSearch is unreachable from a worker.

`task_revoked_handler` is the sharp one: retry.py deliberately resubmits while the old task may still be running, so when that container finally dies its `job-revoked` write would land on the retried attempt's live doc at the same `_id`, and the reaper never sees it because it only touches `job_failed`. The job-lock contention path no longer writes a status doc under a payload a newer attempt owns; it emits a `job_lock_contention` event and raises. `offline_jobs` is the fourth supervisory writer: its redis pair expires at a day and nothing clears it on revoke, so after a retry the old attempt's keys still say job-started while a newer attempt owns the payload.

### A repair daemon for what is already out there

`scripts/reap_orphaned_job_failed.py`, on mozart. An orphan is a `job_failed` doc whose payload also has a doc in a dated `job_status-*` index owned by a later attempt: a different task uuid and a higher `job.retry_count`. retry.py stamps retry_count on every resubmit, so that comparison is monotonic where a timestamp guess is not.

The scan starts from the retried attempts, not from `job_failed`. Retried jobs are rare next to failures (36 dated docs with `retry_count >= 1` in two days on a forward venue, against 1,808 job_failed docs; POP1 carries about 85k), so scanning job_failed and probing per doc would re-run tens of thousands of searches every sweep. Scanning candidates and probing job_failed by id costs two requests per 500, and mget is realtime, so the probe cannot be stale. Results are paired to candidates by `_id`, not position.

Safety properties, each with a test:

- **Deleting is opt-in.** Without `--delete-orphans` the daemon reports what it would delete and deletes nothing. Every PCM overrides `supervisord.conf.mozart` and hand-copies the block, so a flag that had to be *preserved* was protected only by four teams transcribing a comment correctly. Now a block that loses a flag can only get quieter.
- Deletes carry `if_seq_no`/`if_primary_term` from the mget hit. If the retried attempt fails terminally between probe and delete and replaces the doc, the delete 409s and is re-examined next sweep rather than erasing a real failure.
- Redis is checked before reaping; an orphan whose uuid is not terminal there is skipped. A missing key counts as reapable and `HYSDS_JOB_STATUS_EXPIRES` is a day in every shipped celeryconfig, so a window longer than the TTL is refused **when the sweep would delete**, unless `--allow-expired-redis` is given. Reporting over a long window is the documented historical audit and is allowed.
- A terminal failure that exhausted its retry budget is never reaped, because its retry_count is not lower than the candidate's.
- The candidate query targets `job_status-2*`, never `job_status-*`: an index pattern expands aliases, `job_status-current` matches, and that would drag job_failed itself into the scan.
- `--grace-secs` (default 120) keeps the sweep off retries that are still writing, `--lookback-days` (default 1) bounds steady state, `--since` widens it for a historical audit, and `--once` exits non-zero when the sweep threw.
- Reporting sweeps emit the same `job_failed_orphan_reaped` event, flagged `dry_run: true`, so the audit has something to read.
- Consecutive zero-candidate sweeps are warned about, at 10, 100 and 1000.

Every reap records which side of the retry's delete the orphan was indexed on: in the log line, in the event as `mechanism` and `mechanism_basis`, and as tags, since the `event_status` template maps the event body `enabled: false`. The comparison needs no clock. lightweight-jobs v2.1.2 records each swept member's delete position on the resubmitted job as `job_info.retry_delete`, and the same `_id` always routes to the same `job_failed` shard, so the orphan's own `(_primary_term, _seq_no)` compares directly against the `job_failed` entry. `indexed_after_retry_delete` means a late arrival re-created the doc; whether that was the job's own message still in the pipe or a duplicate write is not distinguished, and both are repaired. `indexed_before_retry_delete` is effectively unreachable once the sweep reached `job_failed`, since a not_found delete is still sequenced. Without a mark it falls back to write order under `mechanism_basis: timestamp`, never as a verdict.

`_version` is still emitted but is not the triage. It separated the two mechanisms while the worker wrote the terminal doc twice; the single-write change shifts every version down by one, so a missed delete drops from 2 to 1 and collides with a late write past `index.gc_deletes`.

### Rule evaluation aimed at the right index

Rule evaluation was queued against the dated index recorded in `job_info`. Logstash moves a job-failed doc into `job_failed` as part of the same write, so the evaluation settled on an index the doc had left, exhausted its backoff and never ran. `event_processors._fail_job` and `orchestrator.submit_job` both now pass `index="job_failed"`, which is what `job_worker.py` already passes for a worker-written failure. This is the only change here that causes new job submissions; see the behaviour change below.

`assert_doc_settled` is pinned to the attempt: the uuid is threaded from every `queue_finished_job` call site, the dedup path included. With an unreaped orphan present the probe otherwise passes instantly on the older attempt's doc and the rule decision describes the wrong failure, invisibly, because `retry_job_id` comes from `job.job_info.id`, which is identical across attempts. The settle's inner GET takes `ignore=[404]` and treats absent as not-settled, since the reaper deletes `job_failed` docs on its own schedule.

The per-rule msearch bodies do not get the uuid clause. On this path the evaluated doc lives in `job_failed`, and a write there at the same `_id` replaces any orphan, so there is nothing else for a rule query to match.

## Behaviour change for the release notes

Part D makes user rules evaluate three failure classes for the first time: WorkerLostError, TimeLimitExceeded and ConnectionError. It ships as a code change, so it takes effect on every venue at the next `sds update mozart`, with no config gate.

- OPERA's rule set: every retry rule keys on exit codes 143/255, "credentials", S3 throttling, or DEM and notebook error text. None of those docs carry any of it, so nothing new fires.
- NISAR's: `retry-connection-reset-by-peer` and `retry-connection-closed-by-server` match their phrases anywhere in the doc, which is the text a celery-level ConnectionError task-failed event carries. Both begin firing, with `enable_dedup: false`.

A celery-level `ConnectionError` is not evidence that the job's work failed. The redis status tail is one place a raw ConnectionError escapes `run_job` after the products were published; `is_job_finalized` then reads False because the terminal setex never landed, the ES doc still shows job-started, and `_fail_job` writes job-failed for a job that succeeded. With this change the rules see it, and on NISAR that is an auto-retry that republishes the products.

The per-PCM analysis reads each repo's seed rule files. User rules live in mozart's `user_rules-mozart` index and are runtime-editable through Figaro, so a venue should dump its own live rules before adopting.

If either matters for a given venue, the index choice can be gated on a config key (`index="job_failed" if app.conf.get("EVALUATE_RULES_ON_TASK_FAILED") else res["_index"]`) so a PCM can take the rest without the rule-firing change. Not done here, since HC-648 is staying as one change.

## Three limits worth knowing before planning around the guard

**Two of the six sites are inert wherever a worker cannot reach mozart's OpenSearch.** On a venue whose verdi client verifies certificates, both `get_mozart_es().es.info()` and the guard's `mget` fail in about 0.2 s; `CERTIFICATE_VERIFY_FAILED` was measured on one live venue. The revoked-task handler and the job-lock path then answer UNKNOWN every time and keep their previous unconditional write. Failing open is deliberate — losing the guard beats losing the record — but do not count on live coverage at those two sites until a worker on that venue can make the call.

**It does not separate two first attempts.** `job.retry_count` is written only by retry.py, so a first attempt carries no key and the comparison coerces both sides to 0. With the strict "theirs greater than mine", two different first attempts under one payload — a celery redelivery, or lock contention — neither supersedes the other. That is correct for what this guard is for, since a retry always arrives at N+1, but it means the revoked-task handler is not a defence against redelivery contention.

**`_fail_job` can give up without writing a record.** It raises on both ABSENT and UNKNOWN so its own backoff re-reads, and if the retry's replacement doc is not visible within five tries the backoff exhausts and no terminal record is written. Given that writing during ABSENT would destroy the retried attempt's own doc through logstash's paired delete, losing the record is the better trade, but it is a new way to lose one. `fail_job_gave_up` means either that a record was written and its rules never queued, or that no record was written at all.

## Adoption

**Upgrade order: factotum first.** `queue_finished_job` now sends `index` and `uuid` kwargs on every finished job, and `evaluate_user_rules_job` on a pre-3.3.3 worker raises TypeError on `uuid`. That worker, `[program:user_rules_job]`, runs on the factotum; the producers are on mozart (`event_processors`, `orchestrator`) and verdi (`job_worker`). A venue that rolls mozart and verdi first, the normal order when testing a new PGE container, loses every rule evaluation until the factotum catches up, with no error that points at the upgrade. A new rule worker handed only `index` degrades to the unpinned settle, so the prescribed order is safe at every intermediate state.

**Versions.** lightweight-jobs v2.1.2 or later, or no retry writes a mark and every reaper classification falls back to `mechanism_basis: timestamp`. sdscli 2.1.2, or the same two fields in the PCM's own `job_status` template override.

**The supervisord block.** Added to `configs/supervisor/supervisord.conf.mozart` at `--lookback-days 1` and without `--delete-orphans`. Every PCM overrides this template, so the block has to be added to each one at adoption. hysds-quick-start writes its own config from scratch and does not get the block in this release.

**Turning deleting on.** A reporting run of at least one full day whose would-reap set, read from the `job_failed_orphan_reaped` events carrying `dry_run: true`, matches the venue's own orphan audit exactly, with nothing in it the venue wants kept.

**A historical audit with `--since`.** Size it with `job_failed/_count` over the window first. Measured read-only: OPS-FWD 1,110 `job_failed` docs in total (1,091 in 30 days) and 2,045 retried candidates in 7 days; OPS-POP1 429 in total, 16 in 30 days, no candidates in 7 days.

## Validation

Everything below was run against the final code on a dev venue with mozart and the factotum both on 3.3.3, OpenSearch 3.6.0, and the reaper installed through supervisord.

### One terminal write, A/B on the same probes

The retry rule was disabled so the records stayed put, six probes run on each build:

```
stock worker    _version histogram: {2: 6}
patched worker  _version histogram: {1: 6}     unchanged 20 s later
```

### The upgrade order, demonstrated rather than argued

With mozart on 3.3.3 and the factotum still on 3.3.2, the failure handler queued its rule evaluation and the old rule worker rejected it:

```
mozart   _fail_job queued rule evaluation with a uuid kwarg
factotum Task hysds.task_worker.run_task[2a269ada] raised unexpected:
         TypeError("evaluate_user_rules_job() got an unexpected keyword argument 'uuid'")
```

`task_worker.run_task` splats the payload's kwargs into the function, and 3.3.2's signature is `(job_id, index=None)`.

### The exact classifier, end to end across all three repos

A probe was captured, retried, and its captured message replayed raw into redis after the delete:

```
retry recorded its delete positions on the resubmitted job:
  [{"index": "job_status-1999.01.01", "seq_no": 0}, {"index": "job_status-2026.09.02", "seq_no": 7},
   {"index": "job_failed", "seq_no": 2, "primary_term": 1}]
late write re-created the record at _seq_no 3 / primary_term 1   ->  strictly after the mark

shipped command, no flags:
  DRY-RUN would reap payload_id=…
  sweep summary: {'scanned': 10, 'would_reap': 1, 'by_mechanism': {'indexed_after_retry_delete': 1}}
  orphan still present afterwards: True

--delete-orphans:
  Reaped orphaned job_failed doc: payload_id=…
  sweep summary: {'scanned': 10, 'reaped': 1, 'by_mechanism': {'indexed_after_retry_delete': 1}}
  orphan present afterwards: False

events: dry_run=True then dry_run=False, both basis=seq_no,
        orphan_seq_no 15/3 against a recorded mark of 14/3
```

The non-orphan retried attempts in the same sweeps were left alone. A second retry of the same payload replaced the mark (`job_failed` seq_no 2 to 5), so a stale mark cannot invert a later verdict. A long window is refused when deleting and allowed when reporting.

### The four guard states, on the live cluster

```
venue with no job_failed index and no dailies:
  a payload nobody has written        -> absent
  an index name that is not a daily   -> absent
  no index given at all               -> absent
with a record present:
  we are the newer attempt            -> owned
  we are an older attempt             -> superseded
  we are newer still (higher count)   -> owned
both homes closed mid-probe           -> unknown        <- not absent
reopened                              -> absent
```

The UNKNOWN row is the one that matters: a transport fault reads differently from a deletion, which is what stops a rolling restart from dropping every failure record in flight.

### Rule evaluation

All 8 probes of a batch failed, retried and completed, and the settle against `job_failed` succeeded in about 7 s rather than exhausting its backoff. Separately, a deliberately malformed rule query showed what an un-ignored error costs on this path: the msearch and its fallback each burned ten tries, 68 s in total.

### Under real spot reclaims

Probes that would otherwise exit zero were run on 100%-spot workers, with the job container sent the SIGTERM a reclaim's notice produces, once every three minutes for 90 minutes. A reclaimed job reports `Got non-zero exit code: 255`, one of the two codes the field's retry rule keys on:

```
28 reclaims landed on a running job     25 retry jobs triggered
 0 orphans, sampled every five minutes  0 fail_job_gave_up
 0 job_lock_contention                  job_failed count flat
```

Across those 25 reclaim-driven retries, and a separate four-hour run of 292 retries at higher volume on the same spot workers, `fail_job_gave_up` never fired. The supervisory writers never had to decline a write either, so the guard's verdict counters stayed at zero; the four states were exercised directly instead.

### Response shapes, on both OpenSearch versions in the fleet

Bare 2.11.1 and 3.6.0, identical results:

```
doc GET via a multi-index alias   -> 400 illegal_argument_exception
mget, nonexistent daily           -> per-doc error index_not_found_exception
mget, closed daily                -> per-doc error index_closed_exception
delete: not_found                 -> carries _seq_no and _primary_term
delete: stale if_seq_no           -> 409 version_conflict_engine_exception
_version: create/delete/re-create -> 1, 2, 3 inside gc_deletes; 1 past it
```

The nonexistent-daily result is the one the whole design rests on: the guard probes up to 63 homes and most will not exist on any venue.

### Read-only on the live clusters

INT-FWD runs OpenSearch 3.6.0; OPS-FWD and OPS-POP1 run 2.9.0.

```
                              INT-FWD      OPS-FWD      OPS-POP1
alias GET on job_status-current  400          400          400
64-home _mget                    36-39 ms     146-360 ms   33-51 ms
alias members / dailies          9 / 8        63 / 62      19 / 18
dailies older than 62 d aliased  0            30           7
job_failed docs / 7 d candidates 1925 / 187   1110 / 2045  429 / 0
```

The OPS-FWD alias is the case the always-probe-the-caller's-daily rule is for: 30 aliased dailies fall outside a 62-day walk.

### Tests

178 passed, with 29 pre-existing failures in test_triage and test_utils that are identical on develop. The two verdi guard sites and both `offline_orphaned_jobs` guards have no behavioural test; they are call-site guards over a tested function, and the live runs are what covers them. The orchestrator's two `queue_finished_job` sites are checked at the source, since that module cannot be imported without a celeryconfig.

### How to verify this release in the field

Once this and the delete-side fix are on the same mozart, counting leftover records no longer works: the reaper deletes exactly the documents such an audit would count. Read the `job_failed_orphan_reaped` events in `event_status-*` instead. `mechanism` says only which side of the retry's delete the orphan was indexed on; read it with the mark. A null `retry_delete_mark` with `mechanism_basis: timestamp` is the lane where the sweep never reached `job_failed`; a non-null mark with `indexed_after_retry_delete` is the lane where the doc landed after the delete.

Not validated here: NISAR's two connection-error rules actually firing on a celery-level ConnectionError. That is reasoned from the rules files, and is the behaviour change to watch on the first NISAR venue to adopt this.

## Release

hysds bumped to 3.3.3. CHANGELOG entry added, naming the new deleting daemon, the guard and its three limits, the rule-evaluation behaviour change, the factotum-first upgrade order and the lightweight-jobs and sdscli version dependencies.

[HC-648]: https://hysds-core.atlassian.net/browse/HC-648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HC-640]: https://hysds-core.atlassian.net/browse/HC-640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pymonger:
## Live validation artifacts

Captured from the dev venue used for validation (3-node OpenSearch, real logstash, real celery workers). The venue has since been torn down, so the raw output is recorded here. 79 probe jobs were run in total.

### Part A — the double write is gone

A stock worker writes the job-failed doc twice. Both writes captured by polling the realtime GET:

```
_version=1   uuid=65ce3ed0  status=job-failed  @timestamp=2026-09-02T03:26:17.187Z
_version=2   uuid=65ce3ed0  status=job-failed  @timestamp=2026-09-02T03:26:17.211Z
fields that DIFFER between the two writes: ['@timestamp']
```

24 ms apart, byte-identical otherwise. With this branch's `job_worker.py` installed on the worker, the same probe stays at `_version=1`. Counted across every failed probe on the venue, from the workers' own `log_job_status` output:

```
writes-per-failed-probe histogram: {2: 45, 1: 30}
(2 = stock double write; 1 = with the single-write fix on the worker)
```

45 payloads wrote twice on the stock worker; 30 wrote once on the patched one.

### Part B — a real write-after-delete orphan, found and repaired

Manufactured with the injection recipe: capture the job-failed doc, let the retry delete it and the retried attempt complete at `retry_count=1`, then replay the captured message raw into redis → logstash.

```
retry done: job_failed doc gone; candidate job_status-2026.09.02 status=job-completed uuid=d321d08b retry_count=1
doc fixed up: job_info.index -> job_status-1999.01.01
injected late job-failed for 98fa122b-52c3-4710-b73a-0ff2cbe733b0
ORPHAN re-created in job_failed: True  _version= 4  uuid= 3b0c0c47
  candidate still intact: job_status-2026.09.02 status=job-completed uuid=d321d08b rc=1
```

Dry run — 42 candidates scanned, exactly one flagged, nothing deleted:

```
DRY-RUN would reap payload_id=98fa122b-52c3-4710-b73a-0ff2cbe733b0 orphan_uuid=3b0c0c47-276c-486b-abbd-97bee220d125
  newer_uuid=d321d08b-5369-4a26-83b4-906973101008 retry_count=0->1 orphan_status=job-failed
  orphan_ts=2026-09-02T02:54:59.770Z _version=4
sweep summary: {'scanned': 42, 'would_reap': 1, 'reaped_by_version': {4: 1}}
```

Live run, then verification:

```
Reaped orphaned job_failed doc: payload_id=98fa122b-... orphan_uuid=3b0c0c47-... newer_uuid=d321d08b-...
  retry_count=0->1 orphan_status=job-failed _version=4
sweep summary: {'scanned': 42, 'reaped': 1, 'reaped_by_version': {4: 1}}

orphan still in job_failed: False
candidate intact: job_status-2026.09.02 status=job-completed uuid=d321d08b rc=1
reaped events in event_status: 1
   {"payload_id": "98fa122b-52c3-4710-b73a-0ff2cbe733b0", "newer_uuid": "d321d08b-5369-4a26-83b4-906973101008",
    "orphan_uuid": "3b0c0c47-276c-486b-abbd-97bee220d125", "orphan_retry_count": 0,
    "newer_retry_count": 1, "version": 4}
```

The 41 non-orphan retried attempts in the same sweeps were left alone. Note `_version` reads **4**, not 3, while the Part A double write is still in play: two creates precede the delete. After Part A ships it should read 3. Both land in `reaped_by_version`.

### Part C — the guard, and the bug live testing found in it

Driven against a payload owned by a newer attempt (uuid `bbbb…` at `retry_count 1`) with a stale `job-started` doc for the older uuid `aaaa…` in another alias member:

```
=== PATCHED (guard present: True) ===
  is_job_superseded(...): doc in job_failed belongs to bbbb1111-... at retry_count 1,
    ours is aaaa1111-... at 0; a later attempt owns this payload
  fail_job - aaaa1111-...: a later attempt owns payload partc-1111; not writing job-failed.
  after : job_status-2026.09.01 status=job-started uuid=aaaa1111-000
  after : job_failed  status=job-failed uuid=bbbb1111-000
  VERDICT: payload still owned by the NEWER attempt (correct)

=== STOCK (guard present: False) ===
  after : job_status-2026.09.01 status=None uuid=None
  after : job_failed  status=job-failed uuid=aaaa2222-000
  VERDICT: OLD attempt RESURRECTED over the newer one
```

Live truth table against real OpenSearch, after the fix in `ccf558d`:

```
  superseded(P, A=old attempt, rc=0)   = True    expect True
  superseded(P, B=current owner, rc=1) = False   expect False
  superseded(P, A, rc=5 higher)        = False   expect False
```

The middle row is the defect this exercise caught. The first version of the guard treated *any* different uuid as supersession, so it answered `True` for the payload's current owner whenever an older attempt's leftover existed elsewhere in the alias — an orphaned `job_failed` doc, or an unswept `job-revoked` doc. In production that would make a supervisory writer silently drop a genuine failure, on exactly the state the delete-side defect leaves behind. The comparison is now directional on `job.retry_count`, with ties failing open. The unit tests had asserted the old, wrong behaviour and now cover the regression.

### Part D — the index choice, and what it costs

Same fixture, stock vs patched, capturing the argument passed to `queue_finished_job`:

```
PATCHED  (guard=True):  queue_finished_job(index='job_failed')             called=True
STOCK    (guard=False): queue_finished_job(index='job_status-2026.09.02')  called=True
```

Then what each choice does when rule evaluation actually runs against that doc, after logstash has moved it:

```
settle against job_status-2026.09.02  -> RuntimeError: doc not yet search-visible: partd-live-0001  after 77.6s
settle against job_failed             -> OK in 0.0s
Giving up ensure_job_indexed(...) after 10 tries (RuntimeError: doc not yet search-visible)
```

So on `develop` the evaluation burns its entire backoff and gives up: user rules never see WorkerLostError, TimeLimitExceeded or ConnectionError failures at all. With the fix it settles immediately.

### Environment notes worth keeping

`task_acks_late = True` with `task_reject_on_worker_lost = True` means a dead worker child gets its task **requeued, never failed**. `WorkerLostError` therefore cannot be provoked by killing a worker on this configuration — killing the prefork child yields a redelivery, and killing anything in the container's process tree yields a plain `Got non-zero exit code: 137`. `TimeLimitExceeded` reaches the identical `process_events.fail_job` path and was used instead. That same setting also explains a pattern seen on the forward venue — the same task uuid re-running hours later on a different worker — which is redelivery, not a duplicate-doc bug.

### Final state and test suite

```
job_failed docs=19  orphans=0
probe jobs run in total: 79
```

```
151 passed  (29 pre-existing failures in test_triage / test_utils, identical on develop)
```

**Not validated here:** NISAR's two connection-error rules actually firing on a celery-level ConnectionError. That remains reasoned from the rules files, and is the behaviour change to watch on the first NISAR venue to adopt this.

mcayanan:
## Review notes — HC-648

I read this against the current `develop` and verified the central mechanism independently before commenting. `configs/logstash/indexer.conf.mozart` does exactly what the description says: lines 58–70 replace `[job][job_info][index]` with `job_failed` and set `delete_document => true`, and lines 176–178 issue the paired delete against `original_index`. So a job-failed write really is a *move*, and any search-derived address really is stale for a window. The diagnosis holds, and the ABSENT-means-do-not-write reasoning (writing would resurrect the old attempt *and* destroy the retried attempt's fresh doc via that paired delete) follows directly from it.

The overall shape — prevent, repair, explain; `if_seq_no`/`if_primary_term` on the reap; splitting ABSENT from UNKNOWN after the first version failed closed — is what I'd want here. Comments below are about deployment risk and scope disclosure, not about the approach.

### One thing I checked and want to record as cleared

I initially suspected that removing the second terminal write halved the job-drain increment count, since `fail_job()` calls `job_drain_detected()` unconditionally and `WORKER_CONTIGUOUS_FAILURE_THRESHOLD` is unchanged. It does not. `fail_job()` ends in an unconditional `raise WorkerExecutionError`, so the `except`-branch call and the tail `if job_status_json["status"] == "job-failed"` call are mutually exclusive — the except branch raises and the tail is never reached. Exactly one `fail_job()` call per failed job before and after. The duplicate being removed is the second `log_job_status()` (one direct in the `try`, one inside `fail_job` at the default), which is what the description claims. Drain detection is untouched. Noting it so the next reviewer doesn't have to re-derive it.

### Blocking

**1. The `--dry-run` default doesn't actually reach anyone.**

The supervisord block ships with `--dry-run`, but as the description says, every PCM overrides `supervisord.conf.mozart`, so each one hand-copies the block at adoption. The flag's job is to stop a daemon from deleting production failure records, and its enforcement mechanism is four teams each transcribing a comment correctly. A team that drops the flag while transcribing gets no signal — the daemon just starts deleting.

Consider inverting it: refuse to delete unless an explicit affirmative opt-in is present (a `--i-have-reconciled-a-dry-run` flag, or a celeryconfig key), so the *dangerous* mode requires an addition rather than the *safe* mode requiring a preservation. Same end state for a venue that has done the reconciliation, but the failure direction flips from "silently deletes" to "silently doesn't."

**2. Document which races the guard does not close.**

`job.retry_count` is never set outside `retry.py` — nothing in `orchestrator.py` or `job_worker.py` writes it — so a first attempt carries no key and `job_supersession` coerces missing to `0` on both sides. With the strict `theirs > mine`, two *different first attempts* under one payload (celery redelivery, lock contention) both read `0` and neither supersedes the other.

This is correct for the guard's primary purpose: in the `task_revoked_handler` scenario the new attempt carries `retry_count` N+1 against the old attempt's N, so supersession fires as intended. The description acknowledges the tie case for the lock path. My ask is just that the same limitation be stated for `task_revoked_handler` in the release notes, since that's the path the description calls "the sharp one" and an operator could reasonably read the guard as covering redelivery contention too.

**3. Two of the six guards are inert on at least one real venue, and that should be prominent.**

The description reports that from a verdi worker `get_mozart_es().es.info()` fails in 0.2 s with `CERTIFICATE_VERIFY_FAILED` on INT-FWD, for `info` and the guard's `mget` alike. So on any venue whose worker client verifies certificates, `task_revoked_handler` and the lock-contention path answer UNKNOWN every time and fall through to their previous unconditional write.

Failing open is the right call — losing the guard beats losing the record. But it means the "six supervisory writers now check first" framing overstates live coverage on those venues. This is disclosed in the PR body; I'd like it in the release notes and the operator-facing material too, so nobody plans around coverage they don't have.

### Non-blocking

**4. Backoff exhaustion on ABSENT/UNKNOWN is a new record-loss path, not only a monitoring signal.**

`_fail_job` raises on both ABSENT and UNKNOWN so its backoff re-reads, at `max_tries=5, max_value=10`. If the retry's replacement doc takes longer than that to become visible, the backoff exhausts, `fail_job_gave_up` fires, and the terminal record is never written. Given the paired-delete consequence of writing during ABSENT, losing the record is plausibly the better trade — but it *is* a new way to lose a terminal record, and `fail_job_gave_up` is currently presented as a health signal rather than as "a record was written but its rules never queued, or no record was written at all." Worth making explicit in the release notes.

**5. `is_job_superseded()` is dead code.** Only the tests call it. It's public surface that implies a supported entry point; either use it or drop it.

Nothing else. The scan direction (candidates → probe `job_failed` by id, rather than the reverse), pairing mget results by `_id` instead of position, the `job_status-2*` pattern to keep the alias out of the candidate scan, and refusing a window longer than `HYSDS_JOB_STATUS_EXPIRES` are all the right calls and I couldn't fault them.

mcayanan (approved):
Approving. All 5 items from my earlier review are addressed:

1. `--delete-orphans` is now opt-in; the daemon reports-only by default, so a supervisord block that loses a flag when copied to a PCM override can only get quieter, never destructive.
2. Upgrade-order behavior change documented in the CHANGELOG.
3. Verdi-guard inertness (task_revoked_handler and the lock-contention path answering UNKNOWN on TLS-verifying venues) is now called out explicitly.
4. The ABSENT/UNKNOWN backoff-giveup trade is documented as a new (narrow) way to end up with no terminal record, worth investigating whenever `fail_job_gave_up` fires.
5. `is_job_superseded` dead code dropped.

Verified independently rather than taking the diff summary at face value: the logstash paired-delete mechanism this whole fix is built on (`configs/logstash/indexer.conf.mozart`), the `backoff` library's `raise_on_giveup=True` default that makes item 4 real, and the argparse wiring for item 1 (`dry_run=not args.delete_orphans`). No outstanding concerns.

stirlingalgermissen:
@mcayanan @pymonger these AI PR comments are too long. In the future, please make them shorter and more concise. Ideally PR comments are primarily written by you in part to show a strong understanding of the code being checked in

PR #224, opened 2026-09-02
HC-648: Evaluate user rules against job_failed, not the dated index

pymonger:
## Summary

Both writers of a job-failed doc queue user-rule evaluation against the dated index recorded in `job_info`. Logstash moves that doc into `job_failed` as part of the same write, so the evaluation settles on an index the doc has already left.

Split out of [hysds#223](https://github.com/hysds/hysds/pull/223) because it is the only change there that causes new job submissions, and because it reaches every venue ungated through `sdscli`'s template. It should be adopted on its own schedule.

Relates to [HC-648](https://hysds-core.atlassian.net/browse/HC-648).

## What it costs today

Measured on a dev venue, on the same doc, after logstash had moved it:

```
settle against job_status-2026.09.02  -> RuntimeError: doc not yet search-visible, gave up after 77.6s
settle against job_failed             -> OK in 0.0s
```

`ensure_job_indexed` burns its entire backoff and gives up, so on develop user rules never evaluate the failures `process_events` routes through `fail_job`: WorkerLostError, TimeLimitExceeded and ConnectionError.

## Changes

`event_processors._fail_job` and `orchestrator.submit_job` both pass `index="job_failed"`, which is what `job_worker.py` already passes for a worker-written failure.

## Behaviour change

This is the reason the change is on its own. Those three failure classes start being evaluated by user rules.

- OPERA's rule set: every retry rule keys on exit codes 143/255, "credentials", S3 throttling, or DEM and notebook error text. None of those docs carry any of it, so nothing new fires.
- NISAR's: `retry-connection-reset-by-peer` and `retry-connection-closed-by-server` match their phrases anywhere in the doc, which is the text a celery-level ConnectionError task-failed event carries. Both begin firing.

Two cautions for whoever adopts it.

A celery-level `ConnectionError` is not evidence that the job's work failed. The redis status tail is one place a raw ConnectionError can escape `run_job` after the products were published, and in that case `is_job_finalized` reads False (the terminal setex never landed) and the ES doc still shows job-started, so `_fail_job` writes job-failed for a job that succeeded. With this change the rules then see it. On NISAR that is an auto-retry with `enable_dedup: false`, which republishes the products.

The per-PCM analysis above reads the repo's seed rule files. User rules live in mozart's `user_rules-mozart` index and are runtime-editable through Figaro, so a venue should dump its own live rules before adopting rather than trusting the seeds.

If either is a concern for a given venue, the alternative is to gate the index choice on a config key so a PCM can take the rest of HC-648 without the rule-firing change.

## Test plan

- [x] `test_fail_job_proceeds_when_worker_not_finalized` tightened to `assert_called_once_with("p1", index="job_failed")`. It fails on develop, which passes the dated index the search hit came from.
- [x] Verified live on a dev venue, capturing the argument each version passes: patched `queue_finished_job(index='job_failed')`, stock `queue_finished_job(index='job_status-2026.09.02')`.
- [ ] Not verified: NISAR's two connection-error rules actually firing on a real celery-level ConnectionError. Reasoned from the rules files.

`orchestrator.submit_job` has no test file in this repo; the change there is the same one line.

[HC-648]: https://hysds-core.atlassian.net/browse/HC-648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pymonger:
Folding this back into #223 — HC-648 stays as one PR. Both changes (event_processors._fail_job and orchestrator.submit_job) and the tightened test are now on the HC-648 branch, and the behaviour change is documented in that PR body and flagged for the release notes.

PR #225, opened 2026-09-15
HC-651, HC-652: Reap same-uuid job_failed docs left by celery redelivery; dedup a redelivered finished task under job locking

pymonger:
A worker shut down mid-job writes `job-failed` and is killed before it acks, so the broker delivers the same task, same uuid, to another worker, which runs it to completion. The stale `job_failed` doc then sits beside the `job-completed` doc under one payload_id, and the reaper's retried-attempt scan excludes it at four gates: no `retry_count` moved, same uuid, equal counts, and a shared redis key that now reads `job-completed`.

`scripts/reap_orphaned_job_failed.py` gains a second scan. It starts from executions carrying `job.delivery_info.redelivered` that reached `job-completed`, requires the same uuid, and judges the pair by each execution's own `job_info.time_start` / `time_end`: the re-run must have started after the failed execution ended, and a failed doc that postdates the completed doc is never deleted, since it may be the last trace of a completed job whose re-run then failed. It reports `mechanism: redelivered_after_terminal` with basis `time_start`, records both executions' hosts in the event, and accepts `job-completed` for the shared key. Flags, the opt-in delete and `--since` are unchanged, so `--since` also sweeps pairs that already exist.

With `ENABLE_JOB_LOCKING`, `run_job` now dedups a redelivered task whose uuid already reads `job-completed` or `job-deduped` before any lock is taken, as the non-locking branch already did. Without it the redelivery took the lock and ran again, publish found its own dataset already published, the job failed, and logstash's paired delete removed the good `job-completed` doc. The reaper cannot repair that variant, which is why this rides in the same release. The dedup returns without writing a status, since a `job-deduped` write would land under the shared `_id` and clobber the completed doc. A redelivery after `job-failed` still re-runs on purpose.

The locking-branch bug is tracked separately as [HC-652](https://hysds-core.atlassian.net/browse/HC-652); this PR closes both tickets.

## What the reaper does and does not repair
The redelivered scan repairs the failed-then-completed pair and loses no data. The completed-then-failed variant (a re-run of an already completed job) destroys the completed doc through logstash's paired delete, and the reaper cannot repair it because nothing is left to compare against. That variant is what the worker-side dedup prevents; both halves ship together for that reason.

## Implementation notes for reviewers
- The dedup in the locking branch returns without `log_job_status()`. The lock-contention path about 60 lines above writes a `job-deduped` doc; copying that pattern here would put a `job-deduped` doc under the shared `_id`, the paired delete would remove the good `job-completed` doc, and the fix would cause the bug it prevents. `TestLockingBranchDedup` pins this at the source level.
- The check sits before `check_and_break_stale_lock`, not after the lock is acquired. An early return after `start_heartbeat()` and before the `try/finally` would leak the lock for its 600 s expiry with the heartbeat thread still renewing it. The same test pins the ordering.
- The redelivered scan reaps only when the re-run reached `job-completed`; a re-run that ended `job-offline` or `job-revoked` leaves the failure as the more informative record. The time gate (`candidate.time_start > orphan.time_end`) refuses to delete, it does not tag and delete.

## Testing procedures

Three layers, from fastest to most faithful. Each lists what it proves, the exact procedure, what is inspected, and the result.

### 1. Unit tests

**Proves:** the reaper's two scans, the safety gate, the redis rules and counters, and the locking-branch dedup, in isolation.

```
pip install -e '.[test]'
pytest test/test_reap_orphaned_job_failed.py test/test_job_worker.py
```

To show the branch adds no failures to the full suite, run it on develop as well and compare the failure sets:

```
git worktree add /tmp/hysds-develop develop
(cd /tmp/hysds-develop && pytest test 2>&1 | grep '^FAILED' | sed 's/ - .*//' | sort) > develop.txt
pytest test 2>&1 | grep '^FAILED' | sed 's/ - .*//' | sort > branch.txt
comm -23 branch.txt develop.txt        # must print nothing
git worktree remove --force /tmp/hysds-develop
```

The repo-root `celeryconfig.py` is a Jinja template and must never be imported; the test modules mock `hysds.celery`. Run the tests in an environment with the runtime dependencies installed (`prov_es`, `osaka`, `fab-classic`, ...).

**Coverage added:**
- `test/test_reap_orphaned_job_failed.py`, 14 tests: reap with event and tag checks; the redelivered candidate query; same-uuid requirement; same-execution skip (mid-move, or a `job-offline` sibling); the later-execution safety gate refuses and emits nothing; missing `time_start` is unclassified; shared redis key other than `job-completed` blocks the reap (parametrized); expired key still reaps and is counted; dry run; a retried attempt that was then redelivered; both scans in one sweep; `classify_redelivered` on stamps, the `@timestamp` fallback, nothing to compare, and naive stamps read as UTC. The 30 existing tests pass unchanged apart from the query stub now answering each scan separately.
- `test/test_job_worker.py`: `TestRedeliveredTerminalDup` (status table: first delivery never a duplicate and redis is not consulted; `job-completed` and `job-deduped` are duplicates; `job-failed`, `job-started`, `job-offline`, `job-revoked` and a missing key are not) and `TestLockingBranchDedup` (source-level: the dedup precedes `check_and_break_stale_lock`, and returns without `log_job_status` or `fail_job`).

Two boundary tests added after review of edge cases: equal `time_start` and `time_end` stamps are not "later" (the gate is strict, so clock skew can only err toward keeping a failed doc), and a job without `delivery_info` is never a duplicate.

**Result (2026-09-15):** 60 passed in the two suites. The full suite shows the same 29 environment-dependent failures (docker, S3, `test_utils`) on develop and on the branch, and none unique to the branch. CI green.

### 2. End-to-end on a local stack (real worker, real logstash pipeline)

**Proves:** both defects reproduce with real components on develop-equivalent code, and the branch fixes them, without a cluster.

**Stack** (Docker, all services on one network):

| Component | What runs |
|---|---|
| OpenSearch | 3.6.0, security enabled, with the sdscli `job_status`, `event_status`, `task_status` and `worker_status` index templates installed |
| Logstash | OSS 8.9.0 with the OpenSearch output, running `configs/logstash/indexer.conf.mozart` rendered with Jinja. The job-failed move to `job_failed` and the paired delete are the repo's pipeline unchanged; the only edit is `ssl => false` on the redis input |
| Redis | 7, with a TLS port for hysds clients and a plain port for logstash. hysds needs TLS: every `hysds.log_utils` pool passes `ssl_cert_reqs`, which a plain `redis://` connection rejects |
| RabbitMQ | 3.12 |
| Worker | `celery worker --pool=solo` from the checkout under test, config rendered from `configs/celery/celeryconfig.py.tmpl` with `ENABLE_JOB_LOCKING = True` (`task_acks_late` and `task_reject_on_worker_lost` as shipped) |
| Reaper | `scripts/reap_orphaned_job_failed.py --once -g 1 --lookback-days 1` from each checkout |

**How the redelivery is produced.** A wrapper starts the real worker after replacing `hysds.job_worker.queue_finished_job` with a function that SIGKILLs the worker on its first call. `run_job` makes that call immediately after `log_job_status()` of the terminal doc and before the task returns and is acked, which is the window a graceful shutdown hits in production. The broker requeues the unacked message with `redelivered: true` and a second worker receives it. No hysds code is modified; `--pool=solo` keeps the hook in the process that owns the broker connection.

<details><summary>Worker wrapper</summary>

```python
"""Rig worker: the real hysds celery worker, plus one harness hook.

HC651_KILL_ONCE_FILE set and absent: the first call to queue_finished_job --
which run_job makes immediately after log_job_status() of the terminal doc,
and before the task returns and is acked -- SIGKILLs this process. That is the
graceful-shutdown window from the HC-651 reconstruction: terminal doc written,
user-rules trigger not sent, message not acked, broker requeues on connection
loss. --pool=solo so the hook runs in the process that owns the connection.
"""
import logging
import os
import signal
import sys

import hysds.job_worker as jw

_orig = jw.queue_finished_job
KILL_FILE = os.environ.get("HC651_KILL_ONCE_FILE")


def _killing_queue_finished_job(*args, **kwargs):
    if KILL_FILE and not os.path.exists(KILL_FILE):
        with open(KILL_FILE, "w") as f:
            f.write(f"pid={os.getpid()} args={args} kwargs={kwargs}\n")
        logging.getLogger("hc651").warning(
            "HC651 harness: SIGKILL after the terminal write, before the ack"
        )
        for h in logging.getLogger().handlers:
            h.flush()
        os.kill(os.getpid(), signal.SIGKILL)
    return _orig(*args, **kwargs)


jw.queue_finished_job = _killing_queue_finished_job

from hysds.celery import app  # noqa: E402

queue, name = sys.argv[1], sys.argv[2]
app.worker_main(argv=[
    "worker", "--pool=solo", "--loglevel=INFO", "-Q", queue, "-n", name,
    "--without-mingle", "--without-gossip", "--heartbeat-interval=60",
])
```
</details>

**Jobs** are queued exactly like `hysds.orchestrator.submit_job` output: the same `job_json` shape, a `job-queued` status write, and `run_job.apply_async` with an explicit `task_id`. `params._command` points at a script that behaves differently on the first run and on the redelivery (keyed on a marker file):
- `fail-then-complete`: the first run exits 255, the redelivery exits 0. This is the HC-651 pair.
- `complete-then-fail`: the first run exits 0, the redelivery exits 1. This stands in for publish raising `Dataset already exists`, the HC-652 path; everything downstream of the failure (the `job-failed` write, the move, the paired delete) is the real pipeline.

**Procedure for each scenario:**
1. Start worker 1 with the kill hook on its own queue, then submit one job to that queue.
2. Wait for worker 1 to exit, then for the logstash redis list to stay empty for 5 s.
3. Record state: `GET job_failed/_doc/<payload_id>` and `GET job_status-<date>/_doc/<payload_id>` (both return `_seq_no`, `_primary_term` and `_version`), redis `hysds-job-status-<task_id>` and `hysds-task-worker-<task_id>`, the job-lock keys for the payload, and the queue's ready and unacked counts.
4. Start worker 2 without the hook on the same queue. Wait for `run_job[<task_id>]` to succeed or raise in its log, for the queue to drain, and for logstash to drain.
5. Record the same state again, plus the key lines from both worker logs.

The reaper passes run afterwards against the accumulated state.

**Scenarios, expectations and results (2026-09-15):**

| # | Worker code | Scenario | Expected | Observed |
|---|---|---|---|---|
| S1 | branch | `fail-then-complete` | a `job_failed` doc and a `job-completed` doc with the same uuid; redelivered doc has `redelivered: true`; worker 2 does not dedup a `job-failed` redelivery | as expected: worker 2 logged `redelivered_terminal_dup: status:job-failed`, waited out the stale lock, re-ran; failed doc `time_end` 15:46:28.32, completed doc `time_start` 15:47:46.06 |
| S2 | develop | `complete-then-fail` | the redelivery re-runs, fails, and the paired delete destroys the `job-completed` doc (HC-652) | as expected: completed doc gone, `job_failed` doc from worker 2 with `redelivered: true`, redis `job-failed` |
| S2 | branch | `complete-then-fail` | the redelivery is deduped before the lock with no status write; the completed doc is unchanged | as expected: `redelivered_terminal_dup: status:job-completed`, `Encountered duplicate redelivered job`, task succeeded in 0.03 s; completed doc `_seq_no` 5 and `_version` 3 before and after; no `job_failed` doc |
| S3 | n/a | a retry.py-style orphan seeded through `log_job_status` (attempt A fails without `retry_count`, attempt B completes at `retry_count` 1) | still reaped by the existing scan (regression) | as expected |
| R1 | develop reaper | report mode | finds S3 only; cannot see S1 | `scanned` 1, `would_reap` 1 |
| R2 | branch reaper | report mode | finds S3 and S1 | `would_reap` 2: S3 `written_before_retry_queued`/`timestamp`, S1 `redelivered_after_terminal`/`time_start` |
| R3 | branch reaper | `--delete-orphans` | deletes both; completed docs untouched; events emitted | `reaped` 2; S1 completed doc unchanged; `job_failed_orphan_reaped` events tagged by mechanism and basis, with both hosts recorded |
| R4 | branch reaper | `--delete-orphans` again | idempotent | `reaped` 0 |

The S2-develop victim is never a reaper candidate, which confirms the reaper cannot repair the HC-652 variant.

**Observation:** in S2-branch the killed worker's lock keys stay until they expire (600 s TTL; `last_renewed_at` equal to `acquired_at` because the kill came before the first heartbeat). That is the state any killed worker leaves without a redelivery; develop only cleared it by waiting 72 s, breaking it, and re-running destructively.

**Deviations from production:** RabbitMQ 3.12 rather than 3.13; no TLS on the logstash redis input; solo pool; a shell script instead of a PGE, so the real publish raise is not exercised (it needs grq2). The safety gate is not exercised here because within one daily index the paired delete makes a later failed doc and a completed doc unable to coexist; the unit test covers it.

### 3. Full test on a NISAR PCM cluster (in progress)

**Proves:** the change deploys and runs on a production-shaped cluster, the mission's full smoke test still passes, and both fixes behave on real autoscaled workers, real brokers and the deployed reaper daemon.

**Deploying unreleased hysds onto a v6.4.4 cluster.** The PR changes exactly two runtime files over 3.3.3: `hysds/job_worker.py` and `scripts/reap_orphaned_job_failed.py`. A test-only branch of the PCM adds a provisioning hook that:
1. stops provisioning unless the bundled hysds is 3.3.3, so it cannot silently test unpatched code;
2. downloads this PR's head (`3aca48e`) and overlays the two files in mozart's `~/mozart/ops/hysds`, where the reaper daemon runs with `--delete-orphans`;
3. copies them into mozart's `~/verdi/ops/hysds`, which `sds ship` packages into the per-queue worker bundles, and into factotum's editable `~/verdi/ops/hysds`;
4. bind-mounts the bundle's `job_worker.py` over `/root/verdi/ops/hysds/hysds/job_worker.py` in the verdi container (compose and podman launchers).

Step 4 is required because autoscale workers import hysds from inside the verdi image, not from the bundle, and `sds ship` never syncs `~/mozart/ops` into `~/verdi/ops`. Patching mozart alone does not reach the workers.

**Verification procedure:**
1. The provisioning log shows both hook stages.
2. SHA-256 of the two files on mozart, in mozart's verdi bundle source and on factotum equals `git show 3aca48e:<file> | sha256sum`.
3. On an autoscale worker, the container runs the patched file, and the mount is present:
   ```
   docker exec verdi /root/verdi/bin/python -c "import hysds.job_worker as j; print(hasattr(j, 'redelivered_terminal_dup'))"
   docker exec verdi grep -c job_worker.py /proc/self/mountinfo
   ```
   Expected `True` and `1`. Control: the same commands on an unpatched v6.4.4 worker print `False` and `0`.
4. The full forward dev-e2e smoke test passes.

**Targeted redelivery cases.** Workers allow login only as an unprivileged user, so neither the local stack's kill hook nor an iptables ack block is available. Instead each test job arms a watcher inside its own verdi container:
1. From mozart, queue a `run_job` task directly onto a worker queue in the orchestrator's shape. `params._command` runs a script that reads `delivery_info.redelivered` from its own `_job.json`.
2. First execution: sleep, write the watcher into the job dir, start it detached, and exit 255 (`fail-then-complete`) or 0 (`complete-then-fail`).
3. The watcher polls redis for `hysds-job-status-<task_id>` every 1 ms. When it reads the terminal status it waits 2 ms (so the `rpush` that follows the `setex` has gone out) and kills the verdi container through the mounted Docker socket. If the kill beats the ack, the broker requeues the message with `redelivered: true`.
4. On the first node, confirm the verdi container `Exited (137)` and read the watcher log (status seen, kill sent, HTTP response).
5. Detect the redelivery: `task_status` holds `task-received` events for the same uuid from two different worker hostnames.
6. HC-651 case: the redelivered execution completes. Within one reaper sweep (interval 300 s +/- 75 s, grace 120 s) the `job_failed` doc is gone, the `job-completed` doc is untouched, a `job_failed_orphan_reaped` event carries `mechanism:redelivered_after_terminal` and `basis:time_start`, and the reaper log has the `Reaped orphaned job_failed doc` line for the payload.
7. HC-652 case: the second worker's celery log shows `redelivered_terminal_dup: status:job-completed` and `Encountered duplicate redelivered job`; the `job-completed` doc keeps its `_seq_no` and `_version`; no `job_failed` doc exists.
8. The kill races the ack, so an attempt with no second `task-received` is recorded as a lost race and retried with a new job, up to three attempts. Before the terminal status, each attempt also runs the verification command from step 3 on the node executing the job.
   Test jobs fail with exit code 3 and a retried attempt never arms the watcher: the venue's auto-retry rule matches exit codes 143 and 255 with no job-type filter, and a rule-driven retry of a test job would otherwise re-run the script and kill another container.
9. Tear down, then confirm with paginated queries that no instances, auto scaling groups or launch templates remain for the cluster.

**Edge cases.** Each runs on the same cluster after the smoke test, so no smoke job shares a container with a kill.

*Real workers, through the watcher procedure above:*

| Case | Procedure | Expected |
|---|---|---|
| Redelivery after `job-failed`, then success (HC-651) | `fail-then-complete` on an arm64 queue | pair forms; worker 2 logs `redelivered_terminal_dup: status:job-failed` and re-runs; the reaper removes only the failed doc; the gap between the failed `time_end` and the re-run's `time_start` is recorded as margin against clock skew |
| Redelivery after `job-completed` (HC-652) on arm64 | `complete-then-fail` on an arm64 queue | worker 2 dedups before the lock; completed doc `_seq_no` and `_version` unchanged; no `job_failed` doc |
| Same, on x86_64 | `complete-then-fail` on an x86 queue | same as above |
| Redelivery that fails again | `fail-then-fail` | worker 2 re-runs (not deduped); its failure replaces the first under the same `_id` with `redelivered: true`; no dated doc exists, so the reaper has no candidate, and the failure is still present after two sweeps |
| Operator retry right after a deduped redelivery | `complete-then-fail --retry-after`: submit a real lightweight retry of the payload (same job type, queue and kwargs as the venue's retry rule) while the dead first execution's lock may still exist | the retried attempt acquires the lock (breaking the stale one if its heartbeat is gone) and completes; worker lock lines captured |

*Reaper, seeded through the real write path.* `log_job_status` on mozart writes every doc, so logstash's move to `job_failed` and paired delete apply exactly as for real jobs. Failures are written first, then completions, then the redis states some cases need; the deployed daemon (grace 120 s) judges them on its next sweeps, and a checker compares each payload's docs, shared redis key and `job_failed_orphan_reaped` events with the expectation:

| Case | Seed | Expected |
|---|---|---|
| E1 failed doc is the later execution | failed `time_end` T+60, completed redelivered `time_start` T+30, same uuid | kept; `redelivered_skipped_not_later` (the safety gate) |
| E3 shared key says a further execution is running | pair with `time_start` after `time_end`; shared key set to `job-started` | kept on the first sweep, `redelivered_skipped_redis`; after the key is set to `job-completed`, reaped on the next sweep |
| E4 shared key expired | pair; shared key deleted | reaped; `redis_expired` counted |
| E5 retried attempt that was then redelivered | pair where both docs carry `retry_count` 1 | reaped by the redelivered scan; the retried-attempt scan counts `skipped_same_uuid` |
| E6 retry pair whose newer attempt was redelivered | failed uuid A, completed uuid B at `retry_count` 1 with `redelivered: true` | reaped exactly once, by the retried-attempt scan; no conflict or double delete |
| E7 completed doc without `time_start` | pair, completed doc missing the stamp | kept; `redelivered_skipped_unclassified` |
| E9 lone legitimate failure | failed doc only | kept |
| E10 redelivered completion with no failure | completed redelivered doc only | kept; nothing to reap |
| E11 equal stamps | failed `time_end` equals completed `time_start` | kept; `redelivered_skipped_not_later` |

*No false positives under the real workload:* after the smoke test, confirm the reaper's sweep summaries show nothing reaped, no job doc carries `redelivered: true` without a matching test case, every `job-deduped` doc is an orchestrator dedup (payload-hash duplicate) rather than a worker-side redelivery dedup, and no job failed.

<details><summary>Job script and in-container watcher</summary>

```bash
set -u
PY=/root/verdi/bin/python
read r task <<< "$($PY -c 'import json; d=json.load(open("_job.json")); print(d["delivery_info"]["redelivered"], d["task_id"])')"
echo "hc651 mode=@MODE@ redelivered=$r task=$task host=$(hostname)"
if [ "$r" = "False" ]; then
  sleep @FIRST@
  cat > hc651_watcher.py <<'HC651W'
@WATCHER@
HC651W
  setsid nohup $PY hc651_watcher.py "$task" @WANT@ > hc651_watcher.log 2>&1 < /dev/null &
  sleep 1
  if [ @MODE@ = fail-then-complete ]; then echo "first execution fails"; exit 255; fi
  echo "first execution completes"; exit 0
fi
sleep 20
if [ @MODE@ = fail-then-complete ]; then echo "redelivered execution completes"; exit 0; fi
echo "Dataset already exists: simulated publish failure on the re-run" >&2
exit 1
```

`@MODE@`, `@FIRST@` (seconds the first run sleeps) and `@WANT@` (`job-failed` or `job-completed`) are filled in per case, and `@WATCHER@` is replaced with:

```python
import socket, sys, time
from hysds.celery import app
from redis import StrictRedis
task, want = sys.argv[1], sys.argv[2]
r = StrictRedis.from_url(app.conf.REDIS_JOB_STATUS_URL, ssl_cert_reqs="none")
key = "hysds-job-status-" + task
print("watcher armed at %.6f for %s" % (time.time(), want), flush=True)
t0 = time.time()
while time.time() - t0 < 300:
    v = r.get(key)
    if v is not None and v.decode() == want:
        seen = time.time()
        time.sleep(0.002)
        s = socket.socket(socket.AF_UNIX)
        s.connect("/var/run/docker.sock")
        print("status %s seen at %.6f, kill sent at %.6f" % (want, seen, time.time()), flush=True)
        s.sendall(b"POST /containers/verdi/kill HTTP/1.0\r\nContent-Length: 0\r\n\r\n")
        print(s.recv(200).split(b"\r\n")[0].decode(), flush=True)
        sys.exit(0)
    time.sleep(0.001)
print("timeout waiting for " + want, flush=True)
```
</details>

**Status (updated as steps complete):**

| Step | Status |
|---|---|
| Provision with the hook; both stages logged | done 2026-09-15 |
| Two files match `3aca48e` by SHA-256 on mozart, the verdi bundle source and factotum | done 2026-09-15 |
| Autoscale worker container runs the patched `job_worker.py` | done 2026-09-15: `True`, mount present, SHA-256 matches `3aca48e` on an arm64 small-queue worker and an x86_64 sciflo worker |
| Forward dev-e2e smoke test | done 2026-09-15: apply complete (325 resources); `check_pcm` 14 tests, 13 passed, 1 skipped, 0 failures, 0 errors |
| HC-651 case (`fail-then-complete`, arm64) | done 2026-09-15: the watcher saw `job-failed` and sent the kill 2.7 ms later; the container exited 137 about 77 ms after the job's `time_end`, and the broker requeued the unacked message. Worker 2 received the same uuid with `redelivered: true`, logged `redelivered_terminal_dup: status:job-failed`, re-ran and completed. The pair formed (failed doc `redelivered: false`, completed doc `redelivered: true`); the deployed daemon then deleted only the failed doc (guarded on its `_seq_no`), emitted `mechanism:redelivered_after_terminal` / `basis:time_start`, and left the completed doc at `_version` 2. Harness note: killing only the container left an instance the auto scaling group still counted as healthy, so the requeued message waited for a consumer until the instance was terminated (the driver now terminates it after the kill, which is how production loses it); the 1,013 s gap between the two executions reflects that stall, not production timing |
| HC-652 case (`complete-then-fail`, arm64) | done 2026-09-15: the first execution wrote `job-completed`, its verdi container was killed (exit 137) and its instance terminated through the auto scaling group. The broker redelivered the task to the replacement worker, which logged `redelivered_terminal_dup: status:job-completed` and `Encountered duplicate redelivered job`; the task succeeded in 0.035 s with no lock taken and no status written. The `job-completed` doc kept `_version` 3, `redelivered: false` and its original timestamp; no `job_failed` doc exists; the queue drained. Harness note: `task_status` holds one doc per task and same-id events can be indexed out of order, so it still showed `task-received` after the deduped task succeeded; the driver now judges from the job docs and the worker log |
| HC-652 case (`complete-then-fail`, x86_64) | done 2026-09-15: the first x86 execution wrote `job-completed`; the watcher killed its container 2.8 ms later and the message was requeued (queue showed 1 ready, 0 unacked). After the instance was terminated, the replacement x86 worker logged `redelivered_terminal_dup: status:job-completed` and `Encountered duplicate redelivered job`, succeeding in 0.054 s. The `job-completed` doc kept its uuid, host and `_version` 3, and no `job_failed` doc appeared. One earlier attempt on this queue lost the race (the worker acked 50 ms after the terminal write, before the container died), which is why the driver now detects a lost race from the queue counts and retries |
| Redelivery that fails again (`fail-then-fail`) | done 2026-09-15: the watcher killed the first container 2.5 ms after `job-failed`; the broker requeued the message. Worker 2 logged `redelivered_terminal_dup: status:job-failed` (so no dedup), found the dead first execution's lock unrenewed for 211 s, broke it as stale, acquired it, re-ran and failed again. The second failure replaced the first under the same `_id` (`redelivered: true`, `_version` 2), no dated doc existed, and across two reaper sweeps the payload was never a candidate: no reap, no event, doc unchanged |
| Operator retry right after a deduped redelivery | done 2026-09-15: rerun with the retry submitted 15 s after the redelivery was seen. The redelivery was deduped in 0.042 s, and the dead first execution's lock was still live (both keys with 365 s left). The retry's new attempt found that lock held by the dead task, unrenewed for 237 s, broke it as stale, acquired it and completed in 45 s; the payload's doc is that attempt's `job-completed`, which is what a retry is meant to produce. An earlier run reached the retry only after the lock had expired |
| Seeded reaper edge cases E1 to E11 | done 2026-09-15, all nine as expected. First qualifying sweep: E1 and E11 kept (`redelivered_skipped_not_later` 2), E3 kept (`redelivered_skipped_redis` 1), E7 kept (`redelivered_skipped_unclassified` 1), E9 and E10 kept, E4 reaped (`redis_expired` 1), E5 reaped by the redelivered scan (`skipped_same_uuid` 1), E6 reaped once by the retried-attempt scan; totals `redelivered_scanned` 8, `reaped` 3. After E3's shared key was set to `job-completed`, the next sweep reaped E3 and nothing else, and the kept cases stayed kept |
| No false positives under the smoke workload | done 2026-09-15: 2,964 jobs completed, 0 failed or offline; 0 docs with `redelivered: true`; 3 `job-deduped` docs, all orchestrator payload-hash dedups; 2 retried attempts completed with 0 `job_failed` docs left; 42 reaper sweeps with 0 reaped, 0 would-reap, 0 events |
| Teardown with no leftovers | done 2026-09-15: `Destroy complete! Resources: 325 destroyed.`, destroy.sh exit 0. Paginated queries return zero instances, auto scaling groups, launch templates, lambdas, SQS queues, S3 buckets and volumes for `nisar-jdyoung-0061`, and the `/aws/lambda/nisar-jdyoung-0061-` groups are gone. Two empty `system_logs` log groups (0 stored bytes) were recreated by the CloudWatch agents during instance shutdown, which every prior deploy in this venue also leaves behind. |

### Reproducing the production trigger by hand

On any venue, run a job that sleeps for several minutes, then on its worker run what the self-termination daemons (`spot_termination_detector.py`, `harikiri.py`, `harikiri_sqs.py`) run: `docker stop --time=30 $(docker ps -aq)`, then stop supervisord. The pair forms only when the stop lands after the terminal write and before the ack, so repeat it; with root on the worker, blocking outbound AMQP (port 5671) before the stop makes it deterministic. Inspect with the step 3 commands of section 2.

## Deploy
The reaper change is mozart-only (`sds update mozart` + supervisord restart); the worker change rides the verdi bundle (`sds ship`). Deploy both together. No AMI, sdscli or lightweight-jobs change: `job.delivery_info.redelivered` is already mapped `boolean` in the v6.4.4 `job_status` template.

[HC-652]: https://hysds-core.atlassian.net/browse/HC-652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mcayanan (approved):
Traced the fix end-to-end (redis-key dedup in the job-locking branch, and the two-scan reaper) and ran the full test suite on the PR branch: 60/60 new/modified tests pass, and the rest of the suite is clean (2 pre-existing macOS-only `du`-flag failures unrelated to this change).

Confirmed the logic generalizes correctly beyond a single redelivery:
- Both fixes compare against a single last-write-wins state (the shared redis status key, and the single `job_failed` doc keyed by `payload_id`), so N redeliveries of the same uuid collapse to the same one comparison regardless of N.
- Redelivery interleaved with auto-retry is handled too: the two reaper scans are orthogonal (`_reap_retried` keys off differing uuid + higher retry_count, `_reap_redelivered` keys off same uuid), and `test_a_retried_attempt_that_was_then_redelivered_is_reaped` exercises exactly that interleaving.

One pre-existing gap noted for awareness, out of scope here: the non-locking branch's `redelivered_job_dup()` still has a ~60s window where multiple concurrent redeliveries of a `job-started` task could each independently conclude the original worker is dead and run. Worth a follow-up ticket, not a blocker for this PR.

Approving.

